PR #7, opened 2018-06-20
Add script removing `final` modifier from local variables

dmitrykuzmin:
This PR introduces a script which allows to remove `final` modifier from all local variables in the Java code.

The script takes the directory with Java files as an input and performs refactoring in-place.

The only prerequisite for running the script is having `Python 2.7` installed in the system.

The script is proposed to be located in the new directory `scripts/util`.


dmitrykuzmin:
@armiol PTAL.

PR #8, opened 2018-06-25
Update Dependencies

dmdashenkov:
In this PR we bump the dependency versions.

The major motivation for this is to migrate to Protobuf 3.6.0, which fixes the [deprecated API usage issue](https://github.com/google/protobuf/issues/2054).


dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #9, opened 2018-06-27
Update Gradle Plugins

dmdashenkov:
In this PR we make 3 changes: 
 - introduce the dependency for the AppEngine Gradle plugin;
 - remove the `generate-descriptor-set` plugin;
 - introduce the test output plugin.

The AppEngine plugin is commonly used all in the framework examples and tests, thus should be declared once.

The descriptor set file generation is now configured by the model compiler for out projects and for the Spine clients.

Test output plugin is a configuration of the Gradle `test` task. With its help, we make sure that the failed tests are properly logged. This ad-hoc solution should exist until Gradle has a proper support for the JUnit 5 test output.

Also, in this PR we mark the maps declared in `dependencies.gradle` in order to satisfy the IDEA check.


dmdashenkov:
@armiol, PTAL.

PR #10, opened 2018-07-03
gRPC Dependencies

dmdashenkov:
In this PR we add the Grpc dependencies required for `core-java`.

Also, in this PR we fix a typo in the dependency strings. One string was used 3 times. Now, one of the usages is deleted and one is changed to another value.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -53,7 +53,7 @@ final def versions = [
 
 //noinspection GroovyAssignabilityCheck
 final def build = [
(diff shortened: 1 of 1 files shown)


dmdashenkov:
@armiol, PTAL.

armiol:
@dmdashenkov please pay attention to the correct spelling of things. They are `gRPC`.

dmdashenkov:
@vlad-lubenskyi, FYI. As I remember, you use gRPC dependencies in `users` as well.

PR #18, opened 2018-08-16
Enable `AssignmentToCollectionFieldFromParameter` inspection in Idea

dmitrykuzmin:
This PR enables the `AssignmentToCollectionFieldFromParameter` warning in the common IntelliJ IDEA configuration.

Previously, the check was disabled by default, and the compiler said nothing when the `Collection` was copied by reference from the method parameter. 

Now, it emits a warning and suggests doing deep copy instead.

The private methods are ignored by this check.


dmitrykuzmin:
@alexander-yevsyukov PTAL.

PR #20, opened 2018-09-26
Allow non-static inner classes in tests

dmdashenkov:
In order to avoid warnings on `@Nested` test suites, we allow inner test classes to be non-static even if they could be static.


dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #23, opened 2018-11-01
Remove JUnit 4 dependencies

dmytro-grankin:
This PR removes JUnit 4 and JUnit vintage runner dependencies.

Also, the convention for test methods naming was changed. Now, the names should be in `camelCase` instead of `snake_case`. So, the related inspection profile was enabled.


dmytro-grankin:
@dmdashenkov PTAL.

PR #29, opened 2018-12-11
Support non-Java subprojects by `jacoco.gradle` script

dmytro-grankin:
This PR improves `jacoco.gradle` script.

Now the script will work if the root project or its subprojects aren't Java projects. Previously, it tried to aggregate coverage reports from all subprojects and some of them might be non-Java projects.

Also, if subprojects are configured separately, the script should wait until the evaluation of these subprojects. It is needed to say which of the subprojects are Java projects and it is done using `evaluationDependsOnChildren()`.

### DIFF
--- a/gradle/jacoco.gradle
+++ b/gradle/jacoco.gradle
@@ -20,7 +20,7 @@
 
 // Apply this script to enable JaCoCo test report.
 //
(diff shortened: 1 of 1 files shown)


dmytro-grankin:
@armiol PTAL.

PR #30, opened 2018-12-13
Explicit UTF-8 encoding of source files during generation of javadoc

serhii-lekariev:
This PR adds an option to the `javadoc` gradle task that explicitly states UTF-8 encoding of source files.

This broadens the range of supported environments, since during the execution of the `javadoc` task under some Windows-based environments, encoding different from UTF-8 is assumed, resulting in unrecognized symbols.

### DIFF
--- a/gradle/javadoc-options.gradle
+++ b/gradle/javadoc-options.gradle
@@ -38,6 +38,7 @@ javadoc {
     options.tags = ["apiNote:a:API Note:",
                     "implSpec:a:Implementation Requirements:",
                     "implNote:a:Implementation Note:"]
+    options.encoding = 'UTF-8'
 }
 
 // The below config avoids numerous warnings for missing @param tags.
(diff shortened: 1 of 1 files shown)


gradle/javadoc-options.gradle
```diff
@@ -26,6 +26,9 @@
  * 2. @implSpec - the implementation specification
  * 3. @implNote - the commentaries and notes about the implementation
  *
+ * It also explicitly states the encoding of the source files from which the javadoc is composed,
```
dmdashenkov:
```suggestion
 * It also explicitly states the encoding of the source files from which the Javadoc is composed,
```

gradle/javadoc-options.gradle
```diff
@@ -26,6 +26,9 @@
...
+ * broadening the range of environments that the `javadoc` task is runnable under.
```
dmdashenkov:
```suggestion
 * ensuring correct execution of the `javadoc` task.
```

Otherwise, it seems that there is a documented constraint on the range of environments.

PR #32, opened 2018-12-19
Add Firebase admin to the list of dependencies.

mdrachuk:
This PR adds `firebase-admin` to the list of build dependencies, and bumps `appengineApi` and `httpClient ` versions.


mdrachuk:
@armiol, PTAL.

PR #42, opened 2019-02-08
Prevent side effects of `prepareJsPublication` task

dmytro-grankin:
This PR fixes `prepareJsPulication`. Previously, this task cleaned up the publication directory, so it was not possible to make `prepareJsPublication` depend on other tasks.

The similar problem is described in this [issue](https://github.com/gradle/gradle/issues/1012).


dmytro-grankin:
@YegorUdovchenko PTAL.

PR #47, opened 2019-03-01
Fix PMD crashing on certain files across projects

dmitrykuzmin:
There is a [bug](https://github.com/pmd/pmd/issues/1705) in PMD with nested classes imports resolution.

As a result of this bug, the PMD tries to resolve "dependency of the dependency", which is normally not available on the `compileClasspath`.

In Spine, this leads to modules that depend on `client` gRPC services (`CommandServiceGrpc`, `QueryServiceGrpc`, etc.) having a `pmdMain` task crash.

As a workaround to this issue, we specify that classes from `runtimeClasspath` are now also used during PMD check (but not analyzed themselves, see [doc](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.Pmd.html#org.gradle.api.plugins.quality.Pmd:classpath)).


dmitrykuzmin:
@armiol PTAL.

PR #60, opened 2019-04-19
Show full stacktraces in case of an exception during tests

yuri-sergiichuk:
The new logging configuration prints all necessary exception information directly to the console and thus we do not really need to check the test suite report to find out what happened in the test.


yuri-sergiichuk:
@dmdashenkov PTAL.

PR #69, opened 2019-05-30
Fix generated POM layout.

serhii-lekariev:
This PR fixes the layout of the `pom.xml` file that contains all first-level dependencies.

Example of the `pom.xml` with fixed layout can be found [here](https://gist.github.com/serhii-lekariev/37c1c23b9569c2c3f5147ba429c23f43).


serhii-lekariev:
@armiol, PTAL.

PR #73, opened 2019-06-03
Add `Roaster` dependencies

dmitrykuzmin:
This PR adds `Roaster` library dependencies for the convenient reusage in Spine modules. 

Previously we had a library version, but not the dependencies themselves.

Is a part of https://github.com/SpineEventEngine/base/pull/430.


dmitrykuzmin:
@armiol PTAL.

PR #74, opened 2019-06-06
Extract generic IDEA configurations into a separate script

yuri-sergiichuk:
In the PR I have extracted our commonly-used IDEA configurations into a separate Gradle config script.

The [`org.inferred.processors`](https://github.com/palantir/gradle-processors) plugin is used to automate the configuration of the annotation processors for IDEA.

### DIFF
--- a/gradle/idea.gradle
+++ b/gradle/idea.gradle
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2019, TeamDev. All rights reserved.
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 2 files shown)


gradle/idea.gradle
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * This script performs the default Intellij IDEA configuration:
+ *
+ * <ul>
+ *     <li>applies`idea` plugin;
+ *     <li>applies `org.inferred.processors` plugin;
+ *     <li>sets VCS mappings to Git;
+ *     <li>configures IDEA modules merging behavior;
+ *     <li>enforces javadocs and sources download.
```
dmdashenkov:
```suggestion
 *     <li>enforces Javadocs and sources download.
```

gradle/idea.gradle
```diff
@@ -0,0 +1,64 @@
...
+ * </ul>
+ */
+
+plugins {
+    id 'idea'
+    id 'org.inferred.processors' version '2.2.0'
```
dmdashenkov: This won't work. `plugins {}` closure is only available in `build.gradle` files. Please use the legacy approach here.

yuri-sergiichuk:
I'm closing the PR as I found out, it's not that easy to apply the IDEA plugin properly while it's configuration is set in an external script.
In our case, we need to have 2 separate config scripts for root project and modules plus the configuration itself is not being applied in a way it is expected.

PR #78, opened 2019-07-16
Omit `pom.xml` deletion.

serhii-lekariev:
This PR makes it so the `clean` task of any project that `applies` from the `generatePom` script does not delete the `pom.xml`.

The change is made to match the behavior of the `generateLicenseReport`.


serhii-lekariev:
@armiol, PTAL.

PR #84, opened 2019-08-22
Fix `-proto` artifact publishing

dmdashenkov:
In Gradle v5.6 the behaviour of publication config has been changed. Now it is impossible to have more than one publication with the same group ID, artifact name, and version. This breaks the mechanism of publishing `-proto` artifacts in `core-java`.

In order to work this around, we now publish all the artifacts in a single publication. The artifacts are collected through the `archives` configuration.

This is also more pluggable and helps configure to publish more artifacts in separate projects without changing the `config`.

### DIFF
--- a/gradle/publish-proto.gradle
+++ b/gradle/publish-proto.gradle
@@ -56,6 +56,17 @@ buildscript {
     }
 }
 
+task assembleProto(type: Jar) {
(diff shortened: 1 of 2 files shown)


dmdashenkov:
@armiol, PTAL.

PR #85, opened 2019-08-28
JaCoCo configuration fix

dmdashenkov:
In Gradle of version 5.6, there are some internal changes which break our way of configuring the `jacoco` plugin.

In this PR we fix this config.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL.

PR #86, opened 2019-08-28
Fix Checkstyle suppressions on Windows

dmitrykuzmin:
This PR fixes a minor issue where the Unix-like path separators were used for Checkstyle suppression filter.

Due to this, the intended Checkstyle suppressions didn't work on Windows.

The Checkstyle plugin is only actually used in the `time` project, so the issue remained uncovered for some time.


dmitrykuzmin:
@armiol PTAL.

PR #87, opened 2019-08-29
Extract common Proto JS configuration

dmitrykuzmin:
This PR extracts a common configuration of the JS-based Spine projects to a separate script.

Currently, the Protobuf configuration across `users`, `web` and [`money`](https://github.com/SpineEventEngine/money/pull/14) is almost the same.

This PR extracts the common traits to a new script `configure-proto.gradle`.


dmitrykuzmin:
@armiol PTAL.

PR #91, opened 2019-10-02
Fix `SNAPSHOT` pattern

dmdashenkov:
Since the recent change of versioning conventions, all of our artifacts were published into the "releases" repository due to the check which determined the target repository from the version.

Previously, the requirement was that `SNAPSHOT` versions had `-SNAPSHOT` at the end. This is no longer the case as `SNAPSHOT` versions now may end with a build number (e.g. `1.1.3-SNAPSHOT+1`).

This PR makes the publishing script account for the possibility of build numbers.


dmdashenkov:
@armiol, PTAL.

PR #92, opened 2019-11-06
`EntityState` marker interface

dmitrykuzmin:
This PR extends the default configuration of Spine Model Compiler to apply the `io.spine.base.EntityState` interface to all entity state messages.

See https://github.com/SpineEventEngine/base/pull/494.

Note: `config` with these changes is usable starting from Spine version `1.2.0`.


dmitrykuzmin:
@armiol, @dmdashenkov PTAL.

PR #94, opened 2020-01-28
Nested classes generation configuration

dmitrykuzmin:
This PR is a part of [base#508](https://github.com/SpineEventEngine/base/pull/508).

It enables the generation of strongly-typed columns and messages for all Model Compiler users by default.


dmitrykuzmin:
@armiol, @dmdashenkov PTAL. This is what the configuration looks like after the changes.

PR #98, opened 2020-03-13
Enable NPM audit

dmitrykuzmin:
This PR enables the NPM audit for the JS packages.

The audit level is set to `critical` so the build fails only if critical vulnerabilities are discovered. For all other cases, the audit log is just printed to the console.


dmitrykuzmin:
@armiol PTAL.

PR #100, opened 2020-03-17
Pages publishing script

dmdashenkov:
This PR adds the Gradle script for publishing generated docs and GitHub Pages. The script is moved from `core-java` with a couple of significant changes:
1. The links will now look like `<repo name>/reference/<project name>` instead of `<repo name>/javadoc/<project name>`.
2. The docs for previous versions will be stored under `<repo name>/reference/<project name>/v/<version number>`. The docs for the current version will be both stored under the non-versioned path (as by point 1) and under the versioned path.
3. To add extra docs to publish, the user should append the `ext.generatedDocs` file collection.
4. Extra files are not required anymore, except for the credential file. See the script header for the full preconditions.

### DIFF
--- a/scripts/update-gh-pages.gradle
+++ b/scripts/update-gh-pages.gradle
@@ -1,0 +1,212 @@
+/*
+ * Copyright 2020, TeamDev. All rights reserved.
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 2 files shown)


dmdashenkov:
@armiol, PTAL.

scripts/update-gh-pages.gradle
```diff
@@ -0,0 +1,212 @@
...
+ * pushing commits to the `gh-pages` branch.
+ *
+ * To configure docs publishing, apply this script to the repositories which should generate
+ * the doc:
+ * ```
+ * apply deps.scripts.updateGhPages
+ * ```
+ * After that, the `updateGitHubPages` task will be available for that project.
+ *
+ * By default, the non-internal Javadoc is included into the publication. If need, the publication
```
armiol: `needed`

dmdashenkov:
@armiol, yes, as soon as we migrate in `core-java`.

PR #102, opened 2020-03-18
Fix path to a script

dmdashenkov:
Fixes the subdirectory in which the `update-gh-pages.gradle` script resides.


dmdashenkov:
@armiol, PTAL.

PR #101, opened 2020-03-17
Disable the NPM audit

dmitrykuzmin:
This PR disables the NPM audit task [enabled](#98) previously.

The audit still periodically fails project builds due to service unavailability.

Although the corresonding NPM [issue](https://npm.community/t/enoaudit-from-registry-npmjs-org-503/4642) is closed, the problem still persists.

This PR disables the `auditNodePackages` task so the developers don't have to constantly restart project builds locally and on CI.

PR #103, opened 2020-03-18
Patch audit

### LINKED ISSUE #99: Run NPM packages audit upon `check`, not upon `assemble`
Currently, there's a task `auditNodePackages` which runs package audit. The task is bound to `assemble` (through other tasks). We should instead bind it to check and add a mandatory dependency (`dependsOn`) on the task which constructs the `package-lock.json` file.

dmdashenkov:
The NPM audit sometimes fails. In this PR we:
 - reorder the tasks so that the audit is run upon `check`, not `assemble` (fixes #99);
 - attempt the European NPM registry mirror if the audit fails (fixes #101).

### DIFF
--- a/gradle/js/build-tasks.gradle
+++ b/gradle/js/build-tasks.gradle
@@ -39,6 +39,7 @@ ext {
     }
 }
 
+apply plugin: 'base'
 apply from: deps.scripts.npmCli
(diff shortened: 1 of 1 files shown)


dmdashenkov:
@armiol, PTAL.
@dmitrykuzmin, FYI.

PR #104, opened 2020-03-19
Extract script

dmdashenkov:
This PR extracts a bash script into a separate file.

By default, `eval` function is not supported by the Java process execution mechanism. Thus, we extract it into a separate file and run It with Bash.


dmdashenkov:
@armiol, PTAL.

PR #105, opened 2020-03-20
Fix git reference

dmdashenkov:
In this PR we fix the git repository reference in the `update-gh-pages.gradle`, as well as adding more output if the build fails.

### DIFF
--- a/gradle/update-gh-pages.gradle
+++ b/gradle/update-gh-pages.gradle
@@ -135,14 +135,19 @@ String execute(final String... command) {
     return execute(rootDir, command)
 }
 
+final String repoSlug = System.getenv('TRAVIS_REPO_SLUG')
+if (repoSlug == null || repoSlug.isEmpty()) {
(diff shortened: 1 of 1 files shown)


dmdashenkov:
@armiol, PTAL.

gradle/update-gh-pages.gradle
```diff
@@ -157,6 +149,20 @@ updateGitHubPages.doLast {
...
+    if (repoSlug == null || repoSlug.isEmpty()) {
+        throw new GradleException('`TRAVIS_REPO_SLUG` environmental variable is not set.')
+    }
+
+    /**
+     * The GitHub URL to the project repository.
+     *
+     * <p>The host name is {@code github.com-publish}, though resolved to {@code github.com} by SSH on
+     * Travis due to the SSH configuration. This is used to make it possible to have multiple SSH
+     * private keys on Travis (since there is one already on Travis).
```
armiol: Please clarify.

No one ever will get the original issue from the `(since there is one already on Travis)`. I would describe the original issue first. Then explain why `github.com-publish` is used.

PR #112, opened 2020-05-14
Kotlin Gradle scripts. Part 1 of 2

dmdashenkov:
This PR starts the migration of our Gradle scripts to Kotlin.

The benefits of using Kotlin over Groovy for Gradle scripts are:
 - strict static typing;
 - better IDE support (for IntelliJ IDEA).

For now, we migrate the `deps` object with the whole dependency structure and the `run-build` routines.

The `deps` object declared in `ext` becomes the `Deps` Kotlin `object` type. The `runBuild(..)` function becomes the `RunBuild` class which implements the Gradle `Task` interafce.

Both are declared in the `buildSrc` directory. Now, by pulling the config, a user also copies that directory into the root project directory.

The `buildSrc` directory is treated by Gradle in a special way. It contains a Java project which is built before the target project and all its classes are included into the classpath of the main build.

Now, to use `Deps`, we just need to import it in the script:
```kts
import io.spine.gradle.internal.Deps

dependencies {
    implementation(Deps.build.guava)
}
```

No `apply(..)` statement is required.

To use `RunBuild`, we also import it and specify it as a type for a task:

```kts
import io.spine.gradle.internal.RunBuild

val runIncludedProject by tasks.registering(RunBuild::class) {
    directory = "path/to/project/root"
}
```

For the specifics of using Gradle with Kotlin, see [the official doc](https://docs.gradle.org/current/userguide/kotlin_dsl.html).


dmdashenkov:
@alexander-yevsyukov, @armiol, PTAL.

Note that the old Groovy Gradle scripts are still in place and will remain there until the migration is over in all our repos.

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
```diff
@@ -0,0 +1,303 @@
...
+
+package io.spine.gradle.internal
+
+import org.gradle.api.Project
+import org.gradle.api.artifacts.ConfigurationContainer
+import org.gradle.api.artifacts.dsl.RepositoryHandler
+import java.net.URI
+
+/*
+ * Copyright 2020, TeamDev. All rights reserved.
```
armiol, line 29: A single copyright notice seems to be enough.

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
```diff
@@ -0,0 +1,303 @@
...
+    val hamcrest      = "org.hamcrest:hamcrest-all:1.3"
+    val truth = listOf(
+            "com.google.truth:truth:${Versions.truth}",
+            "com.google.truth.extensions:truth-java8-extension:${Versions.truth}",
+            "com.google.truth.extensions:truth-proto-extension:${Versions.truth}"
+    )
+}
+
+object Scripts {
+    fun testArtifacts(p: Project)          = "${p.rootDir}/config/gradle/test-artifacts.gradle"
```
armiol, line 211: Maybe we could extract the common path portion into a function? Not sure why exactly, but it's now much more text than previously :)

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
```diff
@@ -0,0 +1,303 @@
...
+
+/*
+ * This file describes shared dependencies of Spine sub-projects.
+ *
+ * Inspired by dependency management of the Uber's NullAway project:
+ *  https://github.com/uber/NullAway/blob/master/gradle/dependencies.gradle
+ */
+
+// Repositories to which we may publish. Normally, only one repository will be used.
+// See `publish.gradle.kts` for details of the publishing process.
```
armiol, line 56: This comment block looks misplaced now. Perhaps it should be documenting the `PublishingRepos` object?

dmdashenkov:
@armiol, PTAL again.

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
```diff
@@ -0,0 +1,303 @@
...
+    val ci = "true".equals(System.getenv("CI"))
+    val gradlePlugins = GradlePlugins
+}
+
+object Gen {
+    val javaPoet = "com.squareup:javapoet:${Versions.javaPoet}"
+}
+
+object Grpc {
+    val grpcCore               = "io.grpc:grpc-core:${Versions.grpc}"
```
alexander-yevsyukov, line 159: Why do we need `grpc` prefix assuming it all comes under the `Grpc` object?

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
```diff
@@ -0,0 +1,303 @@
...
+    val grpcStub               = "io.grpc:grpc-stub:${Versions.grpc}"
+    val grpcOkHttp             = "io.grpc:grpc-okhttp:${Versions.grpc}"
+    val grpcProtobuf           = "io.grpc:grpc-protobuf:${Versions.grpc}"
+    val grpcNetty              = "io.grpc:grpc-netty:${Versions.grpc}"
+    val grpcNettyShaded        = "io.grpc:grpc-netty-shaded:${Versions.grpc}"
+    val grpcContext            = "io.grpc:grpc-context:${Versions.grpc}"
+}
+ 
+object Runtime {
+    val floggerSystemBackend = "com.google.flogger:flogger-system-backend:${Versions.flogger}"
```
alexander-yevsyukov, line 169: Can we have nested nested objects? So that it's `Runtime.Flogger.systemBackend`?

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
alexander-yevsyukov: Yes, have a method, please.

dmdashenkov:
@alexander-yevsyukov, PTAL again.

PR #113, opened 2020-05-14
Remove backticks from `pull`

dmdashenkov:
Backticks in `echo` caused the `pull` script to fail. Now we remove them.


dmdashenkov:
@armiol, PTAL.

PR #114, opened 2020-05-18
Comment suppressions

dmdashenkov:
In this PR we add comments on why we suppress some warnings in `deps.kt`.


dmdashenkov:
@armiol, PTAL. Related to [this PR](https://github.com/SpineEventEngine/base/pull/532).

PR #115, opened 2020-05-18
Exclude `protobuf-lite`

dmdashenkov:
It's a commodity in our projects to exclude `protobuf-lite` artifacts from runtime. In this PR we add a method to do just that.


dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #116, opened 2020-05-23
Fall back for failing `cd`

alexander-yevsyukov:
This PR improves the reliability of the copying scripts in case `cd` command fails (e.g. because of absence of the `config` directory).


pull
```diff
@@ -36,7 +36,7 @@
 git submodule update
 
 # Make the `config` dir current.
-cd ./config
+cd ./config || { echo "Error: `config` directory is missing."; exit 1; }
```
yuri-sergiichuk, line 39: The backticks should not be used in shell scripts. They are used as a deprecated way of evaluating variables. (see https://github.com/koalaman/shellcheck/wiki/SC2006)

alexander-yevsyukov: Fixed. Also removed `.bat` file.

PR #117, opened 2020-05-25
Document build script

dmdashenkov:
Here we document the `buildSrc/build.gradle.kts` file specifics.


dmdashenkov:
@armiol, PTAL.

PR #118, opened 2020-05-26
Increment guard

dmdashenkov:
In this PR we implement the `IncrementGuard` plugin.

Before, PRs could override existing versions of Spine libraries intentionally or otherwise. This is a bad practice in general. It also may result in an incompatible change within a single version of the library. This kills any build which depends on that library.

In this PR we introduce the `IncrementGuard` Gradle plugin. The plugin creates the `checkVersionIncrement`  task which fails of the current version of the library has already been published to the Maven repository.

Note that the task checks the version upon all the matching artifacts present in the Maven repository, not only upon the latest one.

The `checkVersionIncrement`  task runs upon `check`.

Fixes [this issue](https://github.com/SpineEventEngine/base/issues/537) in `base`.

### DIFF
--- a/gradle/buildSrc/build.gradle.kts
+++ b/gradle/buildSrc/build.gradle.kts
@@ -26,3 +26,9 @@ repositories {
     mavenLocal()
     jcenter()
 }
+
+val jacksonVersion = "2.11.0"
+
+dependencies {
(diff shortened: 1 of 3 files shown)


dmdashenkov:
@armiol, PTAL.

PR #119, opened 2020-06-01
Less common dependencies

dmdashenkov:
In this PR we add dependencies which usually are only used by one Spine library, such as `web` (e.g. servlet API), `gcloud` (cloud libraries), and `jdbc-storage` (SQL-related dependencies).

From now on, they will be managed centrally, just like all other dependencies.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
After a discussion, we decided to declare non-common dependencies locally, at least for now.

PR #120, opened 2020-06-03
Allow custom compiled scripts

dmdashenkov:
With Kotlin we are able to place script plugins into the `buildSrc` dir. Those plugins become [precompiled](https://docs.gradle.org/current/userguide/custom_plugins.html#sec:precompiled_plugins).

At this point, this is the only usable way in which we can develop custom script plugins which use other Gradle plugins. Just placing a `.gradle.kts` file and applying it via `apply(from = "...")` is flawed, as the file cannot use any type-safe accessors for the Gradle model.

In this repository, we prepare the `buildSrc` project for those plugins.
In library repositories, we may define custom script plugins and place them under `buildSrc/src/main/kotlin`. For example, `custom-jar-magic.gradle.kts`:
```kotlin
plugins {
    java
}

tasks.jar {
    exclude("**/*.properties")
}
```
Then, such plugins become available to the target projects. For example, `spine-lib/build.gradle.kts`:
```kotlin
plugins {
    `custom-jar-magic`
}
```

Later we will migrate the most (or all) the script plugins in this repository (which are currently in Groovy) into precompiled script plugins.

### DIFF
--- a/gradle/buildSrc/build.gradle.kts
+++ b/gradle/buildSrc/build.gradle.kts
@@ -19,8 +19,6 @@
  */
 
 plugins {
-    // Use Kotlin for `buildSrc`.
-    // https://kotlinlang.org/docs/reference/using-gradle.html#targeting-the-jvm
     `kotlin-dsl`
 }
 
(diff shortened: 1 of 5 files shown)


dmdashenkov:
@armiol, PTAL.

gradle/buildSrc/build.gradle.kts
```diff
@@ -21,7 +21,7 @@
 plugins {
     // Use Kotlin for `buildSrc`.
     // https://kotlinlang.org/docs/reference/using-gradle.html#targeting-the-jvm
```
armiol, line 23: Do we still need this?

PR #122, opened 2020-06-13
Fail build on compiler warnings

serhii-lekariev:
This PR forces the build to fail on compiler warnings. This includes Error Prone warnings.

Before, warnings were logged, but the build kept running.

### DIFF
--- a/gradle/javac-args.gradle
+++ b/gradle/javac-args.gradle
@@ -24,7 +24,7 @@
 
 tasks.withType(JavaCompile) {
 
-    if(JavaVersion.current() != JavaVersion.VERSION_1_8) {
+    if (JavaVersion.current() != JavaVersion.VERSION_1_8) {
         throw new GradleException("Spine Event Engine can be built with JDK 8 only." +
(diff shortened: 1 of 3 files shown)


serhii-lekariev:
@armiol, PTAL.

serhii-lekariev:
> As per discussion, let's have this PR merged once we have a PR to `core-java` with the issues addressed.

I've reverted the Error Prone version bump. I suggest that this branch is merged in its current state, and the version is bumped in a follow-up PR, together with the fixes in the existing repositories.

serhii-lekariev:
Let's actually not do that yet, and discuss the warning in `base` first.

gradle/javac-args.gradle
```diff
@@ -33,8 +33,11 @@ tasks.withType(JavaCompile) {
 
     // Explicitly states the encoding of the source and test source files, ensuring
     // correct execution of the `javac` task.
+    //
+    // Also promote compiler warnings to errors, so that build fails on warnings.
```
armiol, line 37: `promotes`.

PR #126, opened 2020-06-20
Specify the -Xmx for the publishing process itself.

yuri-sergiichuk:
This change is an attempt to make sure the publishing process itself could use more memory.


yuri-sergiichuk:
@armiol, PTAL.

PR #128, opened 2020-06-24
Loose increment version check.

yuri-sergiichuk:
Allow repositories that were never published before (and thus have no metadata) to pass the check.

An example here the `pubsub` module I'm currently adding to `gcloud` repository. It was never published before and it's not possible to pass the check now.


gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/CheckVersionIncrement.kt
```diff
@@ -94,9 +95,19 @@ private data class MavenMetadata(var versioning: Versioning = Versioning()) {
...
         }
 
-        fun fetchAndParse(url: URL): MavenMetadata {
-            val metadata = mapper.readValue(url, MavenMetadata::class.java)
-            return metadata
+        /**
+         * Fetches the metadata for the repository and parses the document.
+         *
+         * <p>If the document could not be found, assumes that the module was never
+         * previously released and thus has no metadata.
```
armiol, line 102: Let's kill `previously`.

PR #141, opened 2020-09-17
Run the `CheckVersionIncrement` only during Pull Request builds

### LINKED ISSUE #139: The version increment checking mechanism leads to failed builds when creating git tags
When a git tag is created, e.g. for a new release, it's built on CI like a separate branch. If the target is `master`, the build won't pass due to the `checkVersionIncrement` task failing.

The problem with this is that the GitHub shows this "tag" build status as the status of the last commit to `master`. So, it will look like the commit `Merge pull request ... SpineEventEngine/release-x.x.x` has failed, even when in reality the library was successfully published.

We should probably tweak the `checkVersionIncrement` task so that the misleading information isn't displayed on the GitHub UI.

dmitrykuzmin:
This PR modifies the `IncrementGuard` to run version increment check only during Travis Pull Request jobs.

The `checkVersionIncrement` task will no longer be executed:
1. during local builds;
2. for `merge` commits;
3. when building individual branches;
4. during any other CI checks rather than Travis build.

This change will allow CI checks to pass in situations outside of the "normal" build cycle, for example, when `master` should be re-built after artifacts are already published, or when a git tag is created.

Fixes #139.

### DIFF
--- a/gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/IncrementGuard.kt
+++ b/gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/IncrementGuard.kt
@@ -20,6 +20,7 @@
 
 package io.spine.gradle.internal
 
+import com.google.common.flogger.FluentLogger
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 
@@ -34,6 +35,8 @@ class IncrementGuard : Plugin<Project> {
         const val taskName = "checkVersionIncrement"
     }
 
+    private val logger: FluentLogger = FluentLogger.forEnclosingClass()
+
     /**
      * Adds the [CheckVersionIncrement] task to the project.
      *
@@ -45,6 +48,8 @@ class IncrementGuard : Plugin<Project> {
      */
     override fun apply(target: Project) {
         if (!isTravisPullRequest()) {
+            logger.atInfo().log("The build does not represent a Travis pull request job, " +
+                    "ignoring the `checkVersionIncrement` task.")
             return
         }
         val tasks = target.tasks
--- a/gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/IncrementGuard.kt
+++ b/gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/IncrementGuard.kt
@@ -34,7 +34,17 @@ class IncrementGuard : Plugin<Project> {
         const val taskName = "checkVersionIncrement"
     }
 
+    /**
+     * Adds the [CheckVersionIncrement] task to the project.
+     *
+     * Only adds the check if the project is built on Travis CI and the job is a pull request.
+     *
+     * The task will never run outside of Travis CI or when building individual branches. This is
+     * done to prevent unexpected CI fails when re-building `master` multiple times, creating git
+     * tags, and in other cases that go outside of the "usual" development cycle.
+     */
     override fun apply(target: Project) {
+        if (!isTravisPullRequest()) return
         val tasks = target.tasks
         tasks.register(taskName, CheckVersionIncrement::class.java) {
             repository = PublishingRepos.cloudRepo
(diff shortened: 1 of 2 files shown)


dmitrykuzmin:
@armiol, @yuri-sergiichuk PTAL.

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/IncrementGuard.kt
```diff
@@ -43,4 +55,15 @@ class IncrementGuard : Plugin<Project> {
...
         }
     }
+
+    /**
+     * Returns `true` if the current build is a Travis job which represents a GitHub pull request.
+     *
+     * Implementation note: the `TRAVIS_PULL_REQUEST` environment variable contains the pull
+     * request number rather than `"true"` in positive case, hence the check.
+     */
+    private fun isTravisPullRequest(): Boolean {
```
yuri-sergiichuk, line 65: Can you please add a link to the Travis env vars docs? ( https://docs.travis-ci.com/user/environment-variables/#default-environment-variables )

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/IncrementGuard.kt
```diff
@@ -34,7 +34,19 @@ class IncrementGuard : Plugin<Project> {
...
+     *
+     * Only adds the check if the project is built on Travis CI and the job is a pull request.
+     *
+     * The task will never run outside of Travis CI or when building individual branches. This is
+     * done to prevent unexpected CI fails when re-building `master` multiple times, creating git
+     * tags, and in other cases that go outside of the "usual" development cycle.
+     */
     override fun apply(target: Project) {
+        if (!isTravisPullRequest()) {
+            return
```
yuri-sergiichuk, line 48: Is it possible to add some logs here? To make it really explicit that we're not adding the task on purpose.

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/IncrementGuard.kt
```diff
@@ -34,7 +34,19 @@ class IncrementGuard : Plugin<Project> {
         const val taskName = "checkVersionIncrement"
     }
 
+    /**
+     * Adds the [CheckVersionIncrement] task to the project.
```
yuri-sergiichuk, line 38: Is `[CheckVersionIncrement]` a special notation used to mention Gradle tasks? If not, can we just wrap it into backticks instead?

yuri-sergiichuk (commented):
@dmitrykuzmin overall looks OK to me. The question is how do we test this change? Is there a way reliably test such Gradle things?

@dmdashenkov do you have some experience in this field?

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/IncrementGuard.kt
```diff
@@ -34,7 +34,19 @@ class IncrementGuard : Plugin<Project> {
...
+     *
+     * Only adds the check if the project is built on Travis CI and the job is a pull request.
+     *
+     * The task will never run outside of Travis CI or when building individual branches. This is
+     * done to prevent unexpected CI fails when re-building `master` multiple times, creating git
+     * tags, and in other cases that go outside of the "usual" development cycle.
+     */
     override fun apply(target: Project) {
+        if (!isTravisPullRequest()) {
```
dmdashenkov, line 47: Please disable the task instead of not adding it. This leads to a more transparent and reproducible build.

dmdashenkov (changes requested):

dmdashenkov:
To test the change we apply it to one of our repos and hope check manually. There could be a better version, but automating all things Gradle is painful and should only be attempted when in need.

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/IncrementGuard.kt
dmdashenkov: It's a `{@link }` in Kotlin.

dmitrykuzmin:
@yuri-sergiichuk, @dmdashenkov PTAL again.

dmdashenkov (approved):

armiol (approved):
@dmitrykuzmin LGTM, as soon as Yura's comments are resolved (which I think they are).

yuri-sergiichuk (approved):

PR #148, opened 2020-10-02
Provide "y" as input to `pub publish`

dmdashenkov:
`pub publish` command requires manual user approval. In order for it to pass, we provide the line "y" (as in "yes") to the `pub publish` process via standard input.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #150, opened 2020-10-20
Remove call to deleted method `columns`

dmdashenkov:
Since recent changes in Model Compiler, we no longer have to configure `columns { }` in Gradle. This PR removes the call to that deleted method from `model-compiler.gradle`.


dmdashenkov:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov (approved):
LGTM

Issue #155, opened 2020-11-12
Modify the disclaimer header in each of source files to include a link to Apache 2 License


armiol:
Spine is licensed under Apache 2 License. Its copy is located in the root of each Spine repository.

However, the headers of Spine source files list only the disclaimer text along with the copyright notice. In order to avoid any confusion, we should explicitly mention the Apache 2 license in each of the disclaimer headers. That would allow an arbitrary reader to answer any questions on licensing without having to navigate to the repo root and browsing for the `LICENSE` file.

So, in scope of this issue we should modify the header text to be as follows:

```java
/*
 * Copyright 2020, TeamDev. All rights reserved.
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
 * You may obtain a copy of the License at
 *
 * http://www.apache.org/licenses/LICENSE-2.0
 *
 * Redistribution and use in source and/or binary forms, with or without
 * modification, must retain the above copyright notice and the following
 * disclaimer.
 *
 * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
 * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
 * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
 * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
 * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
 * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
 * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
 * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
 * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
```

The corresponding changes must be done in:

[...]

PR #162, opened 2020-12-11
Update disclaimer header and IDEA copyright template.

yuri-sergiichuk:
This PR updates the copyright disclaimer as per issue #155. From now on, repositories that rely on the `config` module will have the updated disclaimer header.


armiol (approved):
@yuri-sergiichuk LGTM, but please consider to remove the copyright headers from the files, in which they were originally omitted.

PR #168, opened 2021-02-01
Fix the POM generation.

### LINKED ISSUE #170: The POM generation fails on Windows
In the `web` repository the POM generation [fails](https://github.com/SpineEventEngine/web/pull/166/checks?check_run_id=1797600012) with the following stack trace on Windows OS:

```
Caused by: java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java_util_stream_Stream$collect$1.call(Unknown Source)
	at ProjectDependenciesAsXml.projectDependencies(D:\a\web\web\config\gradle\generate-pom.gradle:303)
	at ProjectDependenciesAsXml.of(D:\a\web\web\config\gradle\generate-pom.gradle:270)
	at ProjectDependenciesAsXml$of.call(Unknown Source)
	at ProjectPomXml.projectDependencies(D:\a\web\web\config\gradle\generate-pom.gradle:190)
	at ProjectPomXml.writeTo(D:\a\web\web\config\gradle\generate-pom.gradle:140)
```

yuri-sergiichuk:
Fixes #170.

This PR fixes the POM generation process when there's a dependency with the same name and version.

This usually should not happen if the dependencies are forced and duplicates are excluded, but if not we may have such a situation when we're comparing the name and the version.

### DIFF
--- a/gradle/generate-pom.gradle
+++ b/gradle/generate-pom.gradle
@@ -27,6 +27,8 @@
 import groovy.xml.MarkupBuilder
 import org.gradle.api.internal.artifacts.dependencies.AbstractExternalModuleDependency
 
+import java.util.function.Function
+
 import static java.util.stream.Collectors.toSet
 
 /**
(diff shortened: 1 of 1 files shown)


gradle/generate-pom.gradle
```diff
@@ -472,17 +474,12 @@ class DependencyWithScope implements Comparable<DependencyWithScope> {
...
-            if (this.dependency().group.equals(thatDependency.group)) {
-                if (this.dependency().name.equals(thatDependency)) {
-                    return this.dependency().version.compareTo(thatDependency.name)
-                }
-                return this.dependency().name.compareTo(thatDependency.name)
-            }
-            return this.dependency().group.compareTo(thatDependency.group)
-        }
-        return dependencyPriority() - o.dependencyPriority()
+        return Comparator
```
dmdashenkov, line 477: Please make this `Comparator` a static field in the class.

dmdashenkov (approved):
@yuri-sergiichuk, LGTM with a single comment.

PR #169, opened 2021-02-01
Consolidate dependency declarations

alexander-yevsyukov:
This PR:
  * Encapsulates dependencies so that versions are in the same object with artifact IDs that use them.
  * Forces Kotlin dependencies.
  * Bumps some dependency versions.

### DIFF
--- a/gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
+++ b/gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
@@ -139,8 +139,9 @@ object AppEngine {
     const val gradlePlugin = "com.google.cloud.tools:appengine-gradle-plugin:${gradlePluginVersion}"
 }
 
+// https://checkerframework.org/
 object CheckerFramework {
-    private const val version     = "3.7.1"
+    private const val version = "3.7.1"
     const val annotations = "org.checkerframework:checker-qual:${version}"
     val dataflow = listOf(
         "org.checkerframework:dataflow:${version}",
@@ -148,15 +149,9 @@ object CheckerFramework {
     )
 }
 
-object GradlePlugins {
-    const val errorProne  = ErrorProne.gradlePlugin
-    const val protobuf    = Protobuf.gradlePlugin
-    const val appengine   = AppEngine.gradlePlugin
-    val licenseReport = "com.github.jk1:gradle-license-report:${Versions.licensePlugin}"
-}
-
+// https://errorprone.info/
 object ErrorProne {
-    private const val version = "2.4.0"
+    private const val version = "2.5.1"
     const val gradlePluginVersion = "1.3.0"
     // Taken from here: https://github.com/tbroyer/gradle-errorprone-plugin/blob/v0.8/build.gradle.kts
     const val javacPluginVersion = "9+181-r4173-1"
@@ -173,6 +168,7 @@ object ErrorProne {
     const val gradlePlugin = "net.ltgt.gradle:gradle-errorprone-plugin:${gradlePluginVersion}"
 }
 
+// https://github.com/google/flogger
 object Flogger {
     internal const val version = "0.5.1"
     const val lib     = "com.google.flogger:flogger:${version}"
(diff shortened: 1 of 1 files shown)


gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
```diff
@@ -124,55 +112,189 @@ object Versions {
...
+    const val checkApi = "com.google.errorprone:error_prone_check_api:${version}"
+    const val testHelpers = "com.google.errorprone:error_prone_test_helpers:${version}"
+
+    const val javacPlugin  = "com.google.errorprone:javac:${javacPluginVersion}"
+    const val gradlePlugin = "net.ltgt.gradle:gradle-errorprone-plugin:${gradlePluginVersion}"
+}
+
+// https://github.com/google/flogger
+object Flogger {
+    internal const val version = "0.5.1"
```
dmdashenkov, line 173: Here as well, let's just keep it public.

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
```diff
@@ -74,42 +74,30 @@ object Repos {
...
 object Versions {
-    val checkerFramework = "3.7.1"
-    val errorProne       = "2.4.0"
-    val errorProneJavac  = "9+181-r4173-1" // taken from here: https://github.com/tbroyer/gradle-errorprone-plugin/blob/v0.8/build.gradle.kts
-    val errorPronePlugin = "1.3.0"
-    val pmd              = "6.24.0"
+    val animalSniffer    = "1.19"
+    val apacheHttpClient = "2.1.2"
+    val appengineApi     =  "???"
+    val appenginePlugin  =  "???"
```
dmdashenkov, line 86: What are these versions?

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
```diff
@@ -124,55 +112,189 @@ object Versions {
...
+
+// https://github.com/JetBrains/kotlin
+// https://github.com/Kotlin
+object Kotlin {
+    @Suppress("MemberVisibilityCanBePrivate") // used directly from outside
+    const val version      = "1.4.21"
+    const val reflect      = "org.jetbrains.kotlin:kotlin-reflect:${version}"
+    const val stdLib       = "org.jetbrains.kotlin:kotlin-stdlib:${version}"
+    const val stdLibCommon = "org.jetbrains.kotlin:kotlin-stdlib-common:${version}"
+    const val stdLibJdk8   = "org.jetbrains.kotlin:kotlin-stdlib-jdk8:${version}"
```
dmdashenkov, line 229: If we're going to use the `kotlin` Gradle plugin, we may declare these dependencies as follows:
```kotlin
dependencies {
    implementation(kotlin("stdlib"))
    implementation(kotlin("reflect"))
    implementation(kotlin("<your module here>"))
}
```

The version is still needed to use in the plugin declaration statement.

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
```diff
@@ -74,42 +74,30 @@ object Repos {
...
+ *
+ * For versions of other dependencies please see `version` properties of objects declared below.
+ */
 object Versions {
-    val checkerFramework = "3.7.1"
-    val errorProne       = "2.4.0"
-    val errorProneJavac  = "9+181-r4173-1" // taken from here: https://github.com/tbroyer/gradle-errorprone-plugin/blob/v0.8/build.gradle.kts
-    val errorPronePlugin = "1.3.0"
-    val pmd              = "6.24.0"
+    val animalSniffer    = "1.19"
```
dmdashenkov, line 83: It is a bit odd that some versions are here while others are grouped with the respective artefact classifiers. I guess, we could have a separate object for each dependency group.

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
```diff
@@ -124,55 +112,189 @@ object Versions {
...
+object Build {
+    val animalSniffer = "org.codehaus.mojo:animal-sniffer-annotations:${Versions.animalSniffer}"
+    const val autoCommon = AutoCommon.lib
+    val autoService = AutoService
+    const val appEngine = AppEngine.sdk
+    val checker = CheckerFramework
+    val errorProne = ErrorProne
+    val firebaseAdmin = "com.google.firebase:firebase-admin:${Versions.firebaseAdmin}"
+    val flogger = Flogger
+    const val guava = Guava.lib
```
dmdashenkov, line 282: I would suggest `Deps.guava.lib` instead of `Deps.build.guava`. Or possibly, `Deps.main.guava.lib` or `Deps.prod.guava.lib`, if we want to preserve the grouping of the dependencies into main/test/etc.

dmdashenkov (changes requested):
@alexander-yevsyukov, please see my comments.

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
alexander-yevsyukov: As discussed vocally, let's have more control of visibility so that it's easier to find usage scenarios.

dmdashenkov (approved):

PR #175, opened 2021-02-04
Allow empty commits to `gh-pages`

dmdashenkov:
In this PR we allow the script which automatically updates the generated docs on GitHub Pages to make empty commits to the `gh-pages` branch.

Unless allowed explicitly, empty commits are an error in Git. However, for the purposes of our automatic system, we might want to make such commits. If, for example, nothing but infrastructural files are changed in a PR, or a successful `master` build is restarted, we don't want the build to fail just because the generated reference documentation has not changed.


dmdashenkov:
@armiol, PTAL.

alexander-yevsyukov (approved):
LGTM

PR #176, opened 2021-02-04
Ignore a Flogger-related warning

dmdashenkov:
In this PR we disable the new `FloggerSplitLogStatement` ErrorProne warning.

Splitting Flogger statements is a conscious decision we've made for the sake of simpler logging API with `Logging`. This check will not prevent us from making a mistake. Instead, it is a nuisance.


dmdashenkov:
@alexander-yevsyukov, PTAL.

armiol (approved):

PR #179, opened 2021-03-17
Maven repo credentials without a file

dmdashenkov:
In this PR we allow to specify publishing credentials for a Maven repository via arguments, instead of a properties file.

We also define a preset for GitHub Packages publishing.

### DIFF
--- a/gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
+++ b/gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
@@ -46,10 +46,17 @@ import org.gradle.api.artifacts.dsl.RepositoryHandler
 data class Repository(
     val releases: String,
     val snapshots: String,
-    val credentialsFile: String? = null,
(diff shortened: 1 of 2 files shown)


dmdashenkov:
@armiol, PTAL.

gradle/buildSrc/src/main/kotlin/io/spine/gradle/internal/deps.kt
```diff
@@ -38,10 +40,48 @@ import java.net.URI
...
+ */
 data class Repository(
     val releases: String,
     val snapshots: String,
-    val credentials: String
+    val credentialsFile: String? = null,
+    val credentials: Credentials? = null
+) {
+
+    fun credentials(project: Project): Credentials {
```
armiol, line 53: Could you please comment on the expected form in which the credentials are expected to be laid out in a credentials file? I would see that as a part of the method-level documentation here.

armiol (approved):
@dmdashenkov LGTM with a suggestion.

PR #191, opened 2021-04-14
Use only Kotlin-defined dependencies

alexander-yevsyukov:
This PR eliminates remaining string-based dependency usage in favor of those defined in Kotlin objects under `io.spine.internal.dependency` package.

The `version` property of the `deps` was removed to encourage using Maven coordinates declared in corresponding Kotlin objects.

Kotlin version was bumped to `1.5.0-RC`.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/dependency/AppEngine.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/dependency/AppEngine.kt
@@ -27,10 +27,13 @@
 package io.spine.internal.dependency
 
 // https://cloud.google.com/java/docs/reference
+@Suppress("unused")
 object AppEngine {
     private const val version = "1.9.82"
-    private const val gradlePluginVersion = "2.2.0"
-
     const val sdk          = "com.google.appengine:appengine-api-1.0-sdk:${version}"
-    const val gradlePlugin = "com.google.cloud.tools:appengine-gradle-plugin:${gradlePluginVersion}"
+
+    object GradlePlugin {
+        private const val version = "2.2.0"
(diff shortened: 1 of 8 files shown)


buildSrc/src/main/kotlin/io/spine/internal/dependency/AppEngine.kt
```diff
@@ -27,10 +27,13 @@
...
 object AppEngine {
     private const val version = "1.9.82"
-    private const val gradlePluginVersion = "2.2.0"
-
     const val sdk          = "com.google.appengine:appengine-api-1.0-sdk:${version}"
-    const val gradlePlugin = "com.google.cloud.tools:appengine-gradle-plugin:${gradlePluginVersion}"
+
+    object GradlePlugin {
+        private const val version = "2.2.0"
+        const val lib = "com.google.cloud.tools:appengine-gradle-plugin:$version"
```
yuri-sergiichuk, line 37: In other places we use `{}`. Let's have some consistent style (I prefer with `{}`).
```suggestion
        const val lib = "com.google.cloud.tools:appengine-gradle-plugin:${version}"
```

gradle/dependencies.gradle
```diff
@@ -194,7 +149,6 @@ ext.deps = [
         'gen'      : gen,
         'runtime'  : runtime,
         'test'     : test,
-        'versions' : versions,
```
yuri-sergiichuk, line 197: So everything except this line is backward-compatible. Can we please highlight in the PR description that this is a breaking change and developers should use versions from respective Kotlin objects?

yuri-sergiichuk (approved):
@alexander-yevsyukov LGTM. Please consider suggestions below prior to merging the PR.

PR #194, opened 2021-04-15
Move `.gradle` scripts under `buildSrc`

alexander-yevsyukov:
This PR moves `.gradle` scripts from `config/gradle/` to `buildSrc/src/main/groovy`.

It also eliminates the dependency on directory creation utility from Guava from the build scripts, which currently fails publishing.


buildSrc/src/main/groovy/generate-pom.gradle
```diff
@@ -23,6 +23,7 @@
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
+//file:noinspection GroovyVariableCanBeFinal
```
yuri-sergiichuk, line 26: So we have both this inspection suppression and finals below. Can we please stick to one approach? E.g. disable this check for finals in IDEA config completely?

yuri-sergiichuk (approved):

buildSrc/src/main/groovy/generate-pom.gradle
alexander-yevsyukov: This is a temporary measure to reduce the noise in Groovy files, and in this one in particular. 
In other places, the warning helps to see that variables are not modified in (such a shaky Groovy) code.
Turning off this warning altogether is a bit too much.
As we're going to migrate these scripts to Kotlin, we'd be addressing these things.

PR #206, opened 2021-06-02
Dokka plugin dependency

dmdashenkov:
In this PR we add the Dokka Gradle plugin artifact coordinates.
Note that Dokka uses a different version from Kotlin.

We also remove a doc copied by error in `AnimalSniffer.kt`.


dmdashenkov:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov (approved):
LGTM

PR #212, opened 2021-06-15
Print error message if an included build fails

dmdashenkov:
Now we print the error output of a Gradle build that was launched via `RunBuild` instead of just giving the file name.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #213, opened 2021-06-18
Allow seemingly redundant type qualifiers in Gradle

dmdashenkov:
In some parts of a Gradle script, usage of imports is prohibited.

This is related to how Gradle resolves dependencies for its scripts.
This applies to `buildscript { }` and `plugins { }` blocks.

In this PR, specifically for Gradle files, we turn off the IDEA's inspection which warns us about a redundant use of an FQN where an import could be used.
Unfortunately, we cannot turn off this inspection for a section of a file.


dmdashenkov:
As discussed, we value inspections more than the convenience of not writing suppressions constantly. Thus, closing.

PR #232, opened 2021-08-05
Allow to fetch dependencies from GitHub Packages

armiol:
Previously, GitHub Packages could not be used as a source for dependencies. 

This changeset enables that, making the recently introduced `base-types` — which is only published to GitHub Packages — available to `core` modules.

In future PRs, it is planned to simplify the API and get rid of nasty `project` variable required to read the ausweis.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
@@ -26,11 +26,13 @@
 
 package io.spine.internal.gradle
 
+import io.spine.internal.gradle.PublishingRepos.gitHub
 import java.io.File
 import java.net.URI
 import java.util.*
 import org.gradle.api.Project
 import org.gradle.api.artifacts.dsl.RepositoryHandler
+import org.gradle.api.artifacts.repositories.MavenArtifactRepository
 
 /**
  * A Maven repository.
@@ -170,7 +172,9 @@ object Repos {
 }
 
 /**
- * The function to be used in `buildscript` clauses when fully-qualified call must be made.
+ * Registers the standard set of Maven repositories.
+ *
+ * To be used in `buildscript` clauses when fully-qualified call must be made.
  */
 @Suppress("unused")
 fun doApplyStandard(repositories: RepositoryHandler) {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
```diff
@@ -170,15 +172,49 @@ object Repos {
 }
 
 /**
- * The function to be used in `buildscript` clauses when fully-qualified call must be made.
+ * Registers the standard set of Maven repositories.
+ *
+ * To be used in `buildscript` clauses when fully-qualified call must be made.
```
dmdashenkov, line 177:
```suggestion
 * To be used in `buildscript` clauses when a fully-qualified call must be made.
```

buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
```diff
@@ -170,15 +172,49 @@ object Repos {
...
  */
 @Suppress("unused")
 fun doApplyStandard(repositories: RepositoryHandler) {
     repositories.applyStandard()
 }
 
+/**
+ * Registers the selected GitHub Packages repos as Maven repositories.
+ *
+ * To be used in `buildscript` clauses when fully-qualified call must be made.
```
dmdashenkov, line 187:
```suggestion
 * To be used in `buildscript` clauses when a fully-qualified call must be made.
```

buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
```diff
@@ -170,15 +172,49 @@ object Repos {
...
+}
+
+/**
+ * Applies the repositories hosted at GitHub Packages, to which Spine artifacts were published.
+ *
+ * This method should be used by those wishing to have Spine artifacts published
+ * to GitHub Packages as dependencies.
+ */
+fun RepositoryHandler.applyGitHubPackages(project: Project) {
+    val baseTypes = gitHub("base-types")
```
dmdashenkov, line 203: Let's pass the repository name as an argument to this `fun`. I suspect, more repos will switch to GitHub Packages in the near future.

dmdashenkov (approved):
@armiol, LGTM with a couple of suggestions.

buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
armiol: As discussed, this method will contain more repos from GitHub Packages. Therefore, we won't pass the name to it.

PR #234, opened 2021-08-11
Google Cloud Artifact Registry publishing

dmdashenkov:
### New repositories

In this PR we configure publishing artifacts to and fetching artifacts from the Google Cloud Artifact Registry.
The registry hosts a Maven repo which serves Java artifacts.

Unlike GitHub Packages, Google CAR allows public auth-free access to the artifacts.

Right now, two repositories are created, one for releases and one for snapshots.
In order to fetch the artifacts, consumers would only need to plack the right URL into their build script:
```kotlin
repositories {
    maven {
        url = uri("https://europe-maven.pkg.dev/spine-event-engine/releases")
    }

    maven {
        url = uri("https://europe-maven.pkg.dev/spine-event-engine/snapshots")
    }
}
```

Alternatively, just use `applyStandard()` to use the standard Spine repositories, which include the Google CAR repositories.

In order to publish artifacts, the developers would have to:
 1. Download a service account key. The service account for publishing Maven artifacts has been created in the `spine-event-engine` GCP project.
 2. Place the key file somewhere on the file system and point the `GOOGLE_APPLICATION_CREDENTIALS` env variable to that location.
 3. Add the repository to the publishing repositories:
```kotlin
spinePublishing {
    // ...
    targetRepositories.add(PublishingRepos.cloudArtifactRegistry)
}
```

Note. In order to publish artifacts, one might need to install the Google Cloud CLI tools (`gcloud`) first.

As of now, both repositories are publicly available. Google recommends capping request thoughput. Right now, the quota is set to 100 requests per minute, we can set it higher at any time.

### Also, in this PR...
...we generate more precise artifact names in the license report. Previously, all the artifact IDs contained the `spine-` name prefix. Now, we only include one if the modules are published with the prefix.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
@@ -207,6 +207,11 @@ object Repos {
     val cloudArchive = PublishingRepos.cloudArtifactRegistry.releases
     val cloudArchiveSnapshots = PublishingRepos.cloudArtifactRegistry.snapshots
 
+    @Deprecated(
+        message = "Sonatype release repository redirects to the Maven Central",
+        replaceWith = ReplaceWith("sonatypeSnapshots"),
+        level = DeprecationLevel.ERROR
+    )
     const val sonatypeReleases = "https://oss.sonatype.org/content/repositories/snapshots"
     const val sonatypeSnapshots = "https://oss.sonatype.org/content/repositories/snapshots"
 }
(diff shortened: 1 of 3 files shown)


dmdashenkov:
@armiol, @alexander-yevsyukov, PTAL.

armiol (approved):

buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
```diff
@@ -110,6 +114,35 @@ object PublishingRepos {
...
+     * The experimental Google Cloud Artifact Registry repository.
+     *
+     * In order to successfully publish into this repository, a service account key is needed.
+     * The published must create a service account, grant it the permission to write into
+     * Artifact Registry, and generate a JSON key.
+     * Then, the key must be placed somewhere on the file system.
+     * Env variable `GOOGLE_APPLICATION_CREDENTIALS` must point at the key file.
+     * Once, these preconditions are met, publishing becomes possible.
+     *
+     * Implementation note. Google provides [tools](https://github.com/GoogleCloudPlatform/artifact-registry-maven-tools)
```
alexander-yevsyukov, line 127: This section deserves a heading or a tag.

buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
```diff
@@ -110,6 +114,35 @@ object PublishingRepos {
...
+     * for configuring authentication for the Maven repositories, including a Gradle plugin.
+     * However, the plugin is incompatible with Gradle 7.x at the moment. For now, we reproduce what
+     * plugin does manually. This makes the whole `buildSrc` depend on
```
alexander-yevsyukov, line 130: Please have the definite article before `plugin`.

buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
```diff
@@ -110,6 +114,35 @@ object PublishingRepos {
...
+     * In order to successfully publish into this repository, a service account key is needed.
+     * The published must create a service account, grant it the permission to write into
+     * Artifact Registry, and generate a JSON key.
+     * Then, the key must be placed somewhere on the file system.
+     * Env variable `GOOGLE_APPLICATION_CREDENTIALS` must point at the key file.
+     * Once, these preconditions are met, publishing becomes possible.
+     *
+     * Implementation note. Google provides [tools](https://github.com/GoogleCloudPlatform/artifact-registry-maven-tools)
+     * for configuring authentication for the Maven repositories, including a Gradle plugin.
+     * However, the plugin is incompatible with Gradle 7.x at the moment. For now, we reproduce what
```
alexander-yevsyukov, line 129: Please have a line break before `For now`.

alexander-yevsyukov (approved):
LGTM, with few minor requests on documentation.

PR #237, opened 2021-08-12
Explicitly specify Guava for `buildSrc`

dmdashenkov:
When testing Gradle plugins, we copy `buildSrc` into the test project. This may lead to version clashes.
One such clash occurred after we've added the Google Cloud Artifact Registry tools as a dependency to `buildSrc`. The versions of Guava used in the tool and in our plugins differed.

In this PR we force the newer version of Guava in `buildSrc`.


dmdashenkov:
@yuri-sergiichuk, PTAL.

yuri-sergiichuk (approved):

PR #241, opened 2021-08-31
Run `checkVersionIncrement` for both `releases` and `snapshots`

armiol:
Previously, `checkVersionIncrement` task has been targeting only `releases`. So, release versions could not be overwritten once they had been published. It was still possible to overwrite snapshots.

However, as we started publishing to GitHub Packages, any overwrites became impossible. An attempt to do so now resolves in a HTTP response 409 "Conflict".

This changeset now forces the check to ensure neither of `snapshots` and `releases` repositories contains the version which is to be published.

Please note, that for local builds AND for the CI runs in `master` branches this task will still NOT run. It is a precautious measure, partially explained by the comment in `IncrementGuard`:

```kotlin
    /**
     * Adds the [CheckVersionIncrement] task to the project.
     *
     * Only adds the check if the project is built on Travis CI and the job is a pull request.
     *
     * The task only runs on non-master branches on GitHub Actions. This is done
     * to prevent unexpected CI fails when re-building `master` multiple times, creating git
     * tags, and in other cases that go outside of the "usual" development cycle.
     */
    override fun apply(target: Project) { ... }
```
For now, we still may encounter a need to re-trigger builds in `master` branch if they fail for some unfortunate reason. Therefore, such behaviour is left intact within this PR.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/gradle/CheckVersionIncrement.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/gradle/CheckVersionIncrement.kt
@@ -45,7 +45,7 @@ open class CheckVersionIncrement : DefaultTask() {
     /**
      * The Maven repository in which to look for published artifacts.
      *
-     * We check both the `releases` and `snapshots` repository. Artifacts in either of these repos
+     * We check both the `releases` and `snapshots` repositories. Artifacts in either of these repos
      * may not be overwritten.
      */
     @Input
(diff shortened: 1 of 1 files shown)


armiol:
@alexander-yevsyukov @dmdashenkov PTAL.

dmdashenkov (commented):
LGTM.

dmdashenkov (approved):

PR #247, opened 2021-09-08
Return of the Google Cloud Artifact Registry

dmdashenkov:
In this PR we advance the version of the GCAR cloud library, which fixes the issue with uploading large atrifacts.
This fix allows us, finally, to publish into the GCAR.


alexander-yevsyukov (approved):
LGTM

PR #250, opened 2021-09-16
Version writer

dmdashenkov:
In this PR we introduce `VersionWriter`, a `buildSrc` Gradle plugin that generates a `.properties` file with artifact versions in it. The file is included into the project's JAR.

This is a common routine in our Gradle plugins. Now that we split the plugins into multiple repositories, it's nice to have this part shared via `config` instead of copying it over many Git repos.

Also, in this PR, we update the Protobuf version to `3.18.0`.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/dependency/Protobuf.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/dependency/Protobuf.kt
@@ -29,7 +29,7 @@ package io.spine.internal.dependency
 // https://github.com/protocolbuffers/protobuf
 @Suppress("MemberVisibilityCanBePrivate") // used directly from outside
 object Protobuf {
-    const val version    = "3.17.3"
+    const val version    = "3.18.0"
     val libs = listOf(
         "com.google.protobuf:protobuf-java:${version}",
         "com.google.protobuf:protobuf-java-util:${version}"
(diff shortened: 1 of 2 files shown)


yuri-sergiichuk (approved):
@dmdashenkov LGTM.

One question though, so this new plugin cannot be applied in the `plugins` section as our other plugins, right?
Are there any benefits in writing such plugins comparing to [precompiled script plugins](https://docs.gradle.org/current/userguide/custom_plugins.html#sec:precompiled_plugins)?

dmdashenkov:
@yuri-sergiichuk, we could probably replace the class plugin with a script plugin.

I guess I prefer class plugins when there is a need in a custom task, e.g. `WriteVersions` or `RunBuild`, or a custom extension, e.g. `Publishing`. This gives us the freedom to apply plugins conditionally and/or at a certain point in the build process.

Is there a reason I should feel more inclined towards script plugins?

yuri-sergiichuk:
> @yuri-sergiichuk, we could probably replace the class plugin with a script plugin.
> 
> I guess I prefer class plugins when there is a need in a custom task, e.g. `WriteVersions` or `RunBuild`, or a custom extension, e.g. `Publishing`. This gives us the freedom to apply plugins conditionally and/or at a certain point in the build process.
> 
> Is there a reason I should feel more inclined towards script plugins?

While we do have the freedom you mentioned, such declarations and especially usages of plugins from "certain places" usually lead to issues with Gradle configurations issues and later usages of "afterEvaluate" and "doLast" approach which is generally considered prone to errors and not stable.
Usage in the "plugins" section on the other hand is considered a best practice and allows gradle to evaluate the build graph easier/faster and in a more stable way.
I believe there should be no issues with defining custom tasks from within script plugins.

PR #260, opened 2021-09-27
Fix `VersionWriter`

dmdashenkov:
In this PR we fix the `VersionWriter` so that the generated file is added automatically to the resources.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/gradle/VersionWriter.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/gradle/VersionWriter.kt
@@ -126,11 +126,8 @@ class VersionWriter : Plugin<Project> {
         val task = register("writeVersions", WriteVersions::class.java) {
             versionsFileLocation.convention(project.layout.buildDirectory.dir(name))
             includeOwnVersion()
-            @Suppress("SimpleRedundantLet")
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/io/spine/internal/gradle/VersionWriter.kt
```diff
@@ -126,15 +126,12 @@ class VersionWriter : Plugin<Project> {
         val task = register("writeVersions", WriteVersions::class.java) {
             versionsFileLocation.convention(project.layout.buildDirectory.dir(name))
             includeOwnVersion()
-            @Suppress("SimpleRedundantLet")
-                // Decrease the number of null-safe calls (`?.`).
-            extensions.findByType<JavaPluginExtension>()?.let {
-                it.sourceSets
-                    .getByName("main")
+            project.extensions.findByType<JavaPluginExtension>()!!.apply {
```
armiol, line 129: As discussed, let's extract this as an extension to `Project` (see https://github.com/SpineEventEngine/config/pull/256/files#diff-6fc4e249496a8134f46be7e5e2ccd80e5f4afaf68124f72fd4414341c8856b75R150).

armiol (approved):
@dmdashenkov LGTM with a single modification to make, as per vocal discussion.

PR #266, opened 2021-09-29
Smarter `updateGitHubPages` task

armiol:
This changeset updates the `updateGitHubPages` Gradle task.

Previously, it was skipping all the actions in case the project was a `SNAPSHOT`. However, it led to some issues, as consuming Gradle projects may have had a dependency onto `updateGitHubPages` — whether it was actually registered or not.

Now, we always register `updateGitHubPages` as a Gradle task. But for `SNAPSHOT` projects, it just prints a diagnostic message telling that the actual update is going to be skipped for the `SNAPSHOT`-versioned project.


alexander-yevsyukov (approved):
LGTM

PR #272, opened 2021-10-04
Remove redundant plugin dependency

dmdashenkov:
For some reason, we declared the license report plugin twice: in `buildscript { }` and in `plugins { }`.
Now we only do that in `plugins { }`.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #275, opened 2021-10-06
Add missing credentials

dmdashenkov:
In this PR we add the Gradle plugin portal credentials. The credentials are decrypted before publishing artifacts.

Also, we change the Ubuntu build a workflow to run upon every push, not only in PRs. This allows us to upload code coverage info for non-PR branches (such as `master`) to Codecov.


dmdashenkov:
@armiol, PTAL.

PR #280, opened 2021-10-13
Migrate `jacoco.gradle` to Kotlin

armiol:
This changeset migrates the Groovy version of the JaCoCo configuration to the Kotlin one.

Applied only to multi-project Gradle projects. In order to include the subproject's reports into the aggregated report, they must have `jacoco` plugin applied:

```kotlin

// root project:

JacocoConfig.applyTo(project)

// ...

// In this example, let's consider all subprojects to be Java projects.
subprojects {
    apply {

        // ...

        plugin("jacoco")
    }
}
```

In single-module Gradle project `JacocoConfig` should NEVER be applied, as there is nothing to aggregate. A corresponding exception is thrown in this case.

To produce the coverage reports for a single-module project, one should use a plain `jacoco` Gradle plugin.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/gradle/report/coverage/PathMarker.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/gradle/report/coverage/PathMarker.kt
@@ -67,5 +67,4 @@ internal enum class PathMarker(val infix: String) {
      */
     val length: Int
         get() = this.infix.length
-
 }


buildSrc/src/main/kotlin/io/spine/internal/gradle/report/coverage/FileExtensions.kt
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/**
+ * If this file contains the name of a class — according to the passed [classNameParser]s —
+ * appends the specified [destination] with it.
+ */
+internal fun File.appendTo(
+    destination: MutableList<String>,
+    vararg classNameParser: (file: File) -> String?
+) {
```
dmdashenkov, line 78: Let's change the signature:
```kotlin
internal fun File.matches(vararg matcher: (File) -> String?): List<String>
```

This way we see what's happening a bit better. It's strange to me that the result of the method is stored into a mutable list parameter.

buildSrc/src/main/kotlin/io/spine/internal/gradle/report/coverage/FileFilter.kt
```diff
@@ -0,0 +1,50 @@
...
+ * Utilities for filtering the groups of `File`s.
+ */
+internal object FileFilter {
+
+    /**
+     * Excludes the generated files from this file collection, leaving only those which were
+     * created by human beings.
+     */
+    fun producedByHuman(files: Iterable<File>): Iterable<File> {
+        return files.filter { ! it.isGenerated }
```
dmdashenkov, line 41:
```suggestion
        return files.filter { !it.isGenerated }
```

buildSrc/src/main/kotlin/io/spine/internal/gradle/report/coverage/PathMarker.kt
```diff
@@ -0,0 +1,66 @@
...
+     * of human-produced Java files.
+     */
+    JAVA_OUTPUT_FOLDER("/main/");
+
+    /**
+     * The number of symbols in the marker.
+     */
+    val length: Int
+        get() = this.infix.length
+
```
dmdashenkov, line 65: Please delete this empty line.

PR #282, opened 2021-10-18
Use Maven artifact ID for version files

alexander-yevsyukov:
This PR standardizes the way we treat module names when writing `versions-*.properties` files. Previously, these files were composed using module names of a Gradle project. 

In most of the sub-projects, we also add `spine-` prefix when publishing to Maven (using `spinePublishing/spinePrefix` property). Not taking in account the prefix resulted in the confusion:
  * When adding dependencies in a project, we need to take the prefix into account.
  * But whey they are loaded from the code (when reading `.properties`) we had to “remember” somehow if the artifact had the `spine-` prefix or not.

After this PR, we always Maven artifact ID.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/gradle/VersionWriter.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/gradle/VersionWriter.kt
@@ -26,17 +26,16 @@
 
 package io.spine.internal.gradle
 
+import io.spine.internal.gradle.publish.PublishExtension
 import java.util.*
 import org.gradle.api.DefaultTask
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.file.DirectoryProperty
-import org.gradle.api.plugins.JavaPluginExtension
 import org.gradle.api.provider.MapProperty
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.OutputDirectory
 import org.gradle.api.tasks.TaskAction
-import org.gradle.kotlin.dsl.findByType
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/kotlin/io/spine/internal/gradle/VersionWriter.kt
```diff
@@ -96,15 +108,22 @@ abstract class WriteVersions : DefaultTask() {
...
             properties.store(it, "Dependency versions supplied by the `$path` task.")
         }
     }
+
+    private fun resourceFileName(): String {
+        val artifactId = artifactId()
+        return "versions-${artifactId}.properties"
+    }
+
+    private fun artifactId() = PublishExtension.artifactIdIn(project)
```
dmdashenkov, line 126: Let's make this an extension `val` of the `Project`.

buildSrc/src/main/kotlin/io/spine/internal/gradle/publish/Publish.kt
```diff
@@ -88,12 +85,10 @@ class Publish : Plugin<Project> {
 
     companion object {
         const val taskName = "publish"
-        const val extensionName = "spinePublishing"
     }
 
     override fun apply(project: Project) {
-        val extension = PublishExtension.create(project)
-        project.extensions.add(PublishExtension::class.java, extensionName, extension)
+        val extension = PublishExtension.createIn(project)
```
dmdashenkov, line 91: Same here, this could be an extension `val`.

PR #293, opened 2021-11-09
Fix deprecation message

alexander-yevsyukov:
This PR fixes deprecation message in `slow-tests.gradle` to correctly name the receiver type of `registerTestTasks()` extension function.

Also, this PR fixes KDoc links to types that are not available via fully-qualified references.


armiol (approved):
@yevhenii-nadtochii FYI.

PR #303, opened 2021-11-25
Make `String.isSnapshot()` more readable

dmdashenkov:
We add the name of the optional argument of a method used in the `isSnapshot()` extension method to make it more readable.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #304, opened 2021-11-26
Update CheckStyle plugin to the most recent version

armiol:
Previously, we were using `8.29`. Now, it's `9.1`.


alexander-yevsyukov (approved):
LGTM

PR #313, opened 2021-12-13
Promote using `var` in IDEA

armiol:
This changeset configures IDEA to treat all explicit variable types as warnings.


alexander-yevsyukov (approved):
LGTM

PR #314, opened 2021-12-14
Suppress `SpeadOperator` warning

dmdashenkov:
We add the Sonatype Lift configuration file and suppress Detekt's `SpeadOperator` warning.

As the Lift uses an old version of Detekt, [this bug](https://github.com/detekt/detekt/issues/3145) causes false alarm on any spread operator usage in Kotlin.

For now, we turn the check off. Later, we should re-enable it, once Lift upgrades the Detekt version.

### DIFF
--- a/.lift.toml
+++ b/.lift.toml
@@ -1,0 +1,1 @@
+ignoreRules=["SpreadOperator"]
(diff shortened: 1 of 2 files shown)


dmdashenkov:
@alexander-yevsyukov, PTAL.

armiol (approved):

PR #328, opened 2022-01-17
Custom artifact prefix

alexander-yevsyukov:
This PR:
  * Consolidates script plugins related to publishing under the `io.spine.gradle.publish` package.
  * Adds an ability to set a `customPrefix` value when doing `spinePublishing`.
  * Bumps Protobuf -> `3.19.3`.


armiol (approved):
LGTM.

PR #335, opened 2022-02-18
Merge `Publish` plugin and `PublishExtension` into a single `SpinePublishing` extension

yevhenii-nadtochii:

PR #337, opened 2022-03-16
Use Dokka instead of Javadoc

tripolkaandrey:
This PR changes the tool which we use for generating a documentation artifact. Currently, we are using Javadoc. The PR switches it to [Dokka](https://kotlin.github.io/dokka/1.6.10/). The reason behind this is the better UI/UX of the tool output. Side-by-side comparison of existing UI(Javadoc) vs the new one can be made by visiting the links below:
- [Old](https://spine.io/core-java/reference/client/index.html)
- [New](https://tripolkaandrey.github.io)

The list of things that were also done, except changing the generation of Javadoc JAR to Dokka JAR as a publish artifact, is provided below:
- Bumped version of Dokka;
- Refactored `Dokka` object in `io.spine.internal.dependency` for better readability and following already established code style for Gradle Plugin dependency object;
- Created script plugin to ease adding new functionality to Java project. This plugin configures the path for documentation and adds the necessary plugin for a Java project. As this tool is initially developed for Kotlin projects, without proper setup it will "see Java code as Kotlin one"(for example, the documentation for method `private void demo()` becomes `fun demo()`). More information can be found in this [link](https://github.com/Kotlin/dokka#output-formats).

How created functionality can be used:

```dokka-java.gradle.kts
allprojects {
  apply {
    plugin("dokka-for-java")
  }
}
```
This automatically sets the output folder for this tool as "**buildDirectory**/docs/dokka" and configures dokka for a Java project.

### Important details 

Dokka loads many classes (especially for large multi-module projects), so it relies heavily on the Metaspace. It leads to the OutOfMemory exception on [large projects](https://github.com/awslabs/aws-sdk-kotlin). As suggested by [Dokka developers](https://github.com/Kotlin/dokka/wiki/faq#dokka-throws-javalangoutofmemoryerror-metaspace) and [dev community](https://github.com/Kotlin/dokka/issues/1405#issuecomment-1022089083) you may need to manually increase MaxMetaspaceSize in your `gradle.properties`. This could be done by adding the following line:
```gradle.properties
org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m
```
The 1.6.10 version of Dokka deals with Metaspace much better than previous versions, but there is still a possibility of the exception. For example, [`core-java`](https://github.com/SpineEventEngine/core-java) did not raise any exceptions when using Dokka for generating documentation artifacts for every subproject as part of `publish` task. Some things except mentioned above that could be helpful when dealing with the issue:
- kill the existing Gradle daemon and try again;
- use the `--no-daemon` option when running as a CLI command.

This PR is now on hold because Dokka documentation has a few issues which should be resolved first. The list of open issues can be found below:
- [Dokka generates redundant brackets when processing annotations](https://github.com/Kotlin/dokka/issues/2393)
- [Misleading representation of Java Varargs](https://github.com/Kotlin/dokka/issues/2395)


armiol:
I would say, it's because "UI/UX of the tool _output_" is better.

buildSrc/src/main/kotlin/io/spine/internal/gradle/publish/ArtifactTaskName.kt
```diff
@@ -35,5 +35,6 @@ package io.spine.internal.gradle.publish
 internal enum class ArtifactTaskName {
     sourceJar,
     testOutputJar,
-    javadocJar;
+    javadocJar,
```
armiol, line 38: Another thing here is that today we are going to have [one more PR merged to `config`](https://github.com/SpineEventEngine/config/pull/335). In its scope, this code has been eliminated (probably, for good).

Once that request is merged, I'd like you to perform a back-merge from the `master` branch into this feature branch. And adjust your code accordingly. Let's discuss in person any questions you might have.

armiol (changes requested):
@tripolkaandrey please see my comments.

PR #338, opened 2022-03-23
Add possibility to exclude Google `.proto` files from artifacts

yevhenii-nadtochii:
This PR extract non-declarative code from `base`'s build file.

The extracted code excludes Google `.proto` files from all artifacts.

An example of how to use it:

```
tasks {
    excludeGoogleProtoFromArtifacts()
}
```


yevhenii-nadtochii:
@armiol PTAL

PR #340, opened 2022-03-28
Make publishing of `testJar` artifact optional

yevhenii-nadtochii:
This changeset removes `testJar` from the list of artifacts that are published by default. Publishing of this artifact is optional now and can be enabled via `SpinePublishing.testJar` method.

In addition to that, the PR does the following refactoring:

1. Renames `testOutputJar` task to just `testJar`. Each task within `Artifacts.kt` has a name that matches to pattern `<archive classifier>Jar`. This one was an exception.
2. Extracts `ProtoJar` DSL element for `SpinePublishing` into a separate file.
3. Makes `spinePublishing` not append a hyphen to `artifactPrefix`.
4. Drops usages of `Collection` when it is not required by a domain.


yevhenii-nadtochii:
@armiol @alexander-yevsyukov Please take a look.

alexander-yevsyukov (approved):
LGTM

PR #350, opened 2022-04-04
Force versions of `Flogger.lib` and `Flogger.Runtime.systemBackend`

yevhenii-nadtochii:


yevhenii-nadtochii:
@armiol Please take a look.

buildSrc/src/main/kotlin/io/spine/internal/gradle/DependencyResolution.kt
```diff
@@ -59,7 +62,7 @@ fun doForceVersions(configurations: ConfigurationContainer) {
 /**
  * Forces dependencies used in the project.
  */
-fun ConfigurationContainer.forceVersions() {
```
armiol, line 62: I would leave the previous version of the API as well, but mark it as deprecated. This will make the migration much easier.

armiol (changes requested):
LGTM except for a single comment.

armiol (approved):
LGTM.

PR #351, opened 2022-04-05
Get rid of a circular dependency on `publish` task

yevhenii-nadtochii:
When a root project itself is published (`./gradlew publish`), a circular dependency on `publish` task unveils.

This changeset does the following:

1. Adds a check to prevent a circular dependency on `publish` task.
2. Moves `publish` task accessor from `java.publish` to `publish` package. Those two packages overlapped. Preference was given to just `publish` since it is already used in many places.


yevhenii-nadtochii:
@armiol Please take a look.

armiol (approved):
LGTM.

PR #352, opened 2022-04-12
Extend Dokka support

tripolkaandrey:
## Preface

This PR extends [Dokka](https://kotlin.github.io/dokka/1.6.10/) support for our projects. It should give an option of a smooth migration from Javadoc to Dokka documentation generation. The reason behind this is the better UI/UX of the tool output. A side-by-side comparison of the existing UI(Javadoc) vs the new one can be made by visiting the links below:
- [Old](https://spine.io/core-java/reference/client/index.html)
- [New](https://tripolkaandrey.github.io)

### Dokka documentation as a publish artifact
A new option was added to include `dokkaJar` as an artifact. By default, this option is set to `false`. The artifact can be included using the following setup:
 ```build.gradle.kts
spinePublishing {
     dokkaJar {
         enabled = true
     }
}
```

### Use Dokka in Java projects 

A new script plugin was created to ease adding Dokka to a Java project.
 
Firstly, The plugin configures Dokka for a Java project. As Dokka is initially developed for Kotlin projects, without the proper setup, it "sees Java code as Kotlin one"(for example, the documentation for method `void demo()` becomes 
`fun demo()`). More information can be found [here](https://github.com/Kotlin/dokka#output-formats). 

Secondly, it keeps sure that things annotated with `@Internal` are not present in the resulting documentation. This is done by using the custom Dokka plugin from [`dokka-extensions`](https://github.com/SpineEventEngine/dokka-tools/tree/master/dokka-extensions).

Lastly, things this plugin also takes care of:
- setting the output directory for this tool as "**buildDirectory**/docs/dokka";
- configuring styles and logo to be coherent with [Spine theme](https://spine.io/docs/).
- skipping empty packages;
- displaying inherited members for a type in a separate tab.

How created functionality can be used:

```dokka-java.gradle.kts
allprojects {
  apply {
    plugin("dokka-for-java")
  }
}
```

## Important details 

Dokka loads many classes (especially for large multi-module projects), relying heavily on the Metaspace. It leads to the `OutOfMemory` exception on [large projects](https://github.com/awslabs/aws-sdk-kotlin). As suggested by [Dokka developers](https://github.com/Kotlin/dokka/wiki/faq#dokka-throws-javalangoutofmemoryerror-metaspace) and [community](https://github.com/Kotlin/dokka/issues/1405#issuecomment-1022089083) you may need to manually increase MaxMetaspaceSize in your `gradle.properties`. This could be done by adding the following line:
```gradle.properties
org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m
```
The 1.6.10 version of Dokka deals with Metaspace much better than previous versions, but the exception is still possible. For example, [`core-java`](https://github.com/SpineEventEngine/core-java) did not raise any exceptions when using Dokka for generating documentation artifacts for every subproject as part of the `publish` task. Some things that could also be helpful when dealing with the issue:
- kill the existing Gradle daemon and try again;
- use the `--no-daemon` option when running as a CLI command.


buildSrc/src/main/kotlin/io/spine/internal/gradle/publish/SpinePublishing.kt
```diff
@@ -242,6 +242,28 @@ open class SpinePublishing(private val project: Project) {
...
+     *
+     * ```
+     * spinePublishing {
+     *     includeDokkaJar = true
+     * }
+     * ```
+     *
+     * The resulting artifact is available under "dokka" classifier.
+     */
+    var includeDokkaJar = false
```
armiol, line 265: Let's make this expression syntactically similar to `protoJar` (see above). 

Also, the naming is important. We drop "include" prefix in order to make the whole thing shorter. Imagine if we don't:
```
spinePublishing {
    // ...

    
    includeTestJar { ... }
    includeProtoJar { ... }
    includeDokkaJar ...
}
```
Notice how much space this "include" takes? And the person who reads this needs to traverse down to the meaningful part, skipping "include" every time.

armiol (changes requested):
@tripolkaandrey please see my comment.

armiol (approved):
LGTM.

PR #361, opened 2022-04-25
Fix JDK level

alexander-yevsyukov:
This PR fixes the JDK level (which was altered by IDEA to 1.6 in `misc.xml`, and the change slipped in #360) back to 11.


armiol (approved):
LGTM.

PR #366, opened 2022-04-28
Tune Kotlin compiler flags

alexander-yevsyukov:
This PR:
  1. Adds Kotlin compiler flags that were recently used in ProtoData.
  2. Removes the `X` prefix before `-opt-in`, thus avoiding the warning issued since Kotlin version `1.6.20`.


armiol (approved):
LGTM.

Issue #362, opened 2022-04-26
Avoid duplicated dependencies in `pom.xml`


alexander-yevsyukov:
### The problem
Time after time a `pom.xml` file created when we generate a license report may contain duplicated entries. One of such cases is the build of `tool-base`, which results in the warning issued by code analysis of IDEA:

![image](https://user-images.githubusercontent.com/3116444/165344845-e74c5a52-b5d8-4286-9bf4-2b4186e5ff2c.png)

The dependencies of interest here are:
```
  <dependency>
    <groupId>net.sourceforge.pmd</groupId>
    <artifactId>pmd-java</artifactId>
    <version>6.39.0</version>
  </dependency>
```
and
```
  <dependency>
    <groupId>net.sourceforge.pmd</groupId>
    <artifactId>pmd-java</artifactId>
    <version>6.44.0</version>
  </dependency>
```

The first dependency comes from the usage of the `pmd` plugin in `buildSrc` (for the purpose of the script plugin `pmd-settings.gradle.kts`) The second dependency is what we actually use when applying the `pmd-settings` plugin. It has the following code:

```
pmd {
    toolVersion = Pmd.version
    ...
}
```
`Pmd.version` is the value defined in the dependency object, and it's equal to `6.44.0`.

Doing a similar thing in the `buildSrc/build.gradle.kts` does not help. Build scan still shows the dependency on the version `6.39.0`, which gets down to the dependency report.

We cannot force the version of the `pmd` plugin in the build script because it's the core plugin. Gradle complains, asking to remove the version call.

Forcing the `pmd` dependencies via `resolutionStrantegy.force()` in the `buildSrc/build.gradle.kts` does not help either. It's still `6.39.0` in the build scan.

### Suggested solution
Since we cannot address the version of the PMD, we can probably make the License Report generator to list only the latest version in our dependency reports. Since we use this information only for reference purposes, even when two versions are _really_ used, listing only the most recent won't do the harm. And we'd have one less warning in our workflow in return.

alexander-yevsyukov:
@armiol, FYI.

PR #368, opened 2022-04-29
Filter out duplicated dependencies from `pom.xml`

yevhenii-nadtochii:
This PR addresses #362.

As described in the issue, sometimes we can't overcome the presence of several versions of the same dependency. `pom.xml` is generated upon a whole project and doesn't group dependencies by modules. Specific modules override versions dictated by `config`. And, as a result, several versions of the same dependency leak to `pom.xml`.

`spine-time` repository is an example here. Each subproject is configured to use a specific version of `protoc` from `config`. The configuration is performed from the root project:

```
 protoc {
    artifact = Protobuf.compiler
}
```

Meanwhile, `time` module applies our `mc-java` plugin, and not the last version. The plugin overrides the version of `protoc`, specified by the root. As a result, the module uses its own version.

---

This PR makes `DependencyWriter` filter out duplicates by group and name. Only the one, with the newest version, is retained. The duplicates are logged.

Generation of license reports is not changed, since dependencies there are grouped by modules.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/gradle/report/pom/DependencyWriter.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/gradle/report/pom/DependencyWriter.kt
@@ -111,7 +111,6 @@ fun Project.dependencies(): SortedSet<ScopedDependency> {
     val result = deduplicate(dependencies)
         .map { it.scoped }
         .toSortedSet()
-
     return result
 }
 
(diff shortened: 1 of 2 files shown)


armiol (changes requested):
Please log some details on the duplicates, as we discussed.

yevhenii-nadtochii:
@armiol PTAL

armiol (approved):
LGTM.

PR #370, opened 2022-05-06
Write manifests and improvements for Windows

alexander-yevsyukov:
This PR adds storing of necessary attributes is done by the script plugin `write-manifest.gradle.kts` which should be applied to each subproject which needs version information. It was previously [introduced in `tool-base`](https://github.com/SpineEventEngine/tool-base/pull/21).
 
It also brings stability improvements for running under Windows:
 * `GitHubPackages.kt` was modified to extract the ZIP archive using the [zip4j](https://github.com/srikanth-lingala/zip4j) library, instead of using `unzip` command line utility which is not available in vanilla Windows.
 * `RunGradle.kt` was modified to use the `_out` directory instead of `build`. Using `build` for debug output of processes created by `ProcessBuilder` failed the `clean` command under Windows because of files already opened in the directory.


armiol (approved):
LGTM.

PR #372, opened 2022-05-10
Minor improvements to Dokka

tripolkaandrey:
This PR fixed a few problems with Dokka.

### Excluding documentation about generated code from Dokka's output
Documentation about generated types is missing. We should have documentation about them as it is a part of our public API. This PR fixes the problem.

### Links to the wrong(old) Java documentation
Dokka adds a link to the Java official documentation when it encounters a Java type(`String`, `Integer`, etc.). By default, these links lead to Java 8 documentation. This PR manually sets the Java version to 11 in Dokka's configuration.

### No documentation about `protected` members
Dokka, by default, documents only the `public` code. We should also have information about the `protected` code in the documentation. This PR configures the tool to document `protected` code also. Unfortunately, Dokka does not display access modifiers when the source code is Java. They have a problem with the `kotlin-as-java` plugin. The [сorresponding issue ](https://github.com/Kotlin/dokka/issues/2492) in [Kotlin/dokka](https://github.com/Kotlin/dokka) has been created.

### Bump Dokka
This PR bumps Dokka to `1.6.21`. The security vulnerability of all Dokka packages dependent on `jsoup` should be gone. Also, this version supports Kotlin `1.6.21`.

### Reformat documentation
Spine developers have introduced a new guideline regarding the recommended line length of comments. I partly adjust to this recommendation in this PR.


armiol (approved):
LGTM.

PR #378, opened 2022-06-03
Generate and publish Dokka documentation for releases prior to the `2.0.0` version

tripolkaandrey:
## Preface

This PR introduces a script to generate and publish Dokka documentation for releases of Spine projects prior to the `2.0.0` version. This is required to migrate from the Javadoc to the Dokka tool. Otherwise, It would be strange if a user needed to look at the documentation for an older version of a library using a different UI in some other place. 

## Problem statement

Old projects use an older version of Kotlin and Gradle. The [current](https://github.com/SpineEventEngine/config/blob/master/buildSrc/src/main/kotlin/dokka-for-java.gradle.kts) Dokka configuration uses the latest possible Dokka release (and, therefore, Kotlin). So generating documentation for old releases using a new configuration is, to say at least, problematic. We need to somehow make an old car drive with a new engine or develop an approach that would not require making Frankenstein's monster out of old releases.

## Solution

We cannot just go through every release and inject the current configuration to generate documentation. It would require constant adjusting Dokka configuration to the API available according to the Kotlin version used in a release. This approach would take ages to study every release of Dokka and adjust the configuration. Moreover, it will produce output which is not consistent with the current one. 

Another problematic approach would be upgrading Kotlin and Gradle versions in old projects. It produces so many different and unpredictable errors that only considering it is already a wasted effort.

So a sophisticated approach was designed. Instead of interfering with a release's source files, we could divide documentation generation into two parts. The part where `.class` and other resources are produced(or simply `classes` Gradle task) and the second one where Dokka generates documentation by processing these resources. For the first part, we need to use the release's configuration. Otherwise, we will run into issues. However, we can use any configuration for the second part and do not bother ourselves with making it compatible with a release.

This approach led to creating a Gradle project with a minimum setup to use the current Dokka configuration. We can inject modules into this project. Inside this project, there is the `generate.sh` script. During its run, a repository is cloned into a subfolder, a target release is checked out, and the `classes` task is completed in a module using native to the release configuration. After that, the module with the `classes` task output is copied to the parent folder(Gradle project where Dokka is configured). The only thing remaining is to run the `dokkaHtml` task in this parent folder. This task will skip running the `classes` task as resources are already present and go directly to generating documentation. 

The publishing part is a simple interaction with a repository by committing and pushing generated documentation to the `gh-pages` branch.

## Instructions

Below there are provided instructions for the script. They reflect entirely those in the `scripts/publish-documentation` folder.
___
### Description

This directory contains a script and serves as a workspace for it. The script generates and publishes Dokka documentation for releases of Spine projects prior to the `2.0.0` version. Documentation is published to the `gh-pages` branch of the target repository.

### Important details

- The script works with the repository over HTTPS, so if you are not authenticated in `git`, you will be prompted to do so using a username and a personal access token.
- The script conceptually relies on the fact that each release has a corresponding tag.
- It is assumed that releases of Spine projects prior to the `2.0.0` version use Java 8.
- Documentation is generated using [this configuration](../../buildSrc/src/main/kotlin/dokka-for-java.gradle.kts).
- The script modifies the parent directory during its run. At the end, of its run the script reverts 
  the directory to the initial state.

### Prerequisites

Prerequisites for the target repository:
- be a Gradle project;
- have at least one tag and module.

Prerequisites for running the script:
- have [`git`](https://git-scm.com/downloads) and [`jenv`](https://github.com/jenv/jenv#12-adding-your-java-environment) installed;
- have Java **1.8** installed and [registered](https://github.com/jenv/jenv#12-adding-your-java-environment) in `jenv`;
- have Java **11** installed, registered and [set as global](https://github.com/jenv/jenv#13-setting-a-global-java-version) in `jenv`.

### Usage

The script should be launched from this directory and follow the template provided below:
```Bash
./publish.sh repositoryUrl='' tags='x,y,*z' modules='x,y,z'
```

Description of parameters:
* `repositoryUrl` - a GitHub HTTPS URL.
* `tags` - a list of comma-separated git tags. A tag marked with an asterisk is considered 'primary'. A tag without it is considered 'secondary'.
* `modules` - a list of comma-separated module names.

An example is provided below:
```Bash
./publish.sh repositoryUrl='https://github.com/SpineEventEngine/core-java.git' tags='v1.7.0,*v1.8.0' modules='core,client'
```

After running the example, the following happens:

- Documentation for the tag `v1.7.0` is generated and pushed to the `gh-pages`. This tag is passed unmarked, so it is considered 'secondary'. It means that changes are published only to the 'v' directory. The following directory structure is produced:
  ```
    /(root)
    └───dokka-reference
    │   │
    │   └───client
    │   │   └───v
    │   │       └───1.7.0
    │   │           │   client
    │   │           │   images
    │   │           │   scripts
    │   │           │   styles
    │   │           │   index.html
    │   │           │   navigation.html
    │   │
    │   └───server
    │       └───v
    │           └───1.7.0
    │               │   server
    │               │   images
    │               │   scripts
    │               │   styles
    │               │   index.html
    │               │   navigation.html
    │
  ```

- Documentation for the tag `v1.8.0` is generated and pushed to the `gh-pages`. This tag is marked with an asterisk, so it is considered 'primary'. It means that changes are made not only to the 'v' directory but also to files/directories alongside it. The following directory structure is produced:
    ```
    /(root)
    └───dokka-reference
    │   │
    │   └───client
    │   │   │   client
    │   │   │   images
    │   │   │   scripts
    │   │   │   styles
    │   │   │   index.html
    │   │   │   navigation.html
    │   │   │
    │   │   └───v
    │   │       └───1.7.0
    │   │       │    │   client
    │   │       │    │   images
    │   │       │    │   scripts
    │   │       │    │   styles
    │   │       │    │   index.html
    │   │       │    │   navigation.html
    │   │       │   
    │   │       └───1.8.0
    │   │           │   client
    │   │           │   images
    │   │           │   scripts
    │   │           │   styles
    │   │           │   index.html
    │   │           │   navigation.html
    │   │
    │   └───server
    │       │   server
    │       │   images
    │       │   scripts
    │       │   styles
    │       │   index.html
    │       │   navigation.html
    │       │
    │       └───v
    │           └───1.7.0
    │           │    │   server
    │           │    │   images
    │           │    │   scripts
    │           │    │   styles
    │           │    │   index.html
    │           │    │   navigation.html
    │           │   
    │           └───1.8.0
    │               │   server
    │               │   images
    │               │   scripts
    │               │   styles
    │               │   index.html
    │               │   navigation.html
    │
    ```

#### Edge cases

- If any of the passed tags already has Dokka documentation published to the `gh-pages` branch, the script tries to overwrite it if there are changes in a tool output. The presence of changes depends on how much the Dokka configuration has changed since the last publication. If an output matches the published one, a commit is not made.

- If no tags are marked 'primary', they all are considered 'secondary' and processed accordingly.

- If multiple tags are marked 'primary', then each of these tags is published as 'primary' one by one. That means the last in the list marked as 'primary' overwrites all others as if they were published as 'secondary'.

### OS details

The script was developed under and for the macOS. It should not have problems working on a Linux distribution. However, it was not meant and tested to do so.


scripts/dokka/README.md
```diff
@@ -0,0 +1,45 @@
...
+Description of parameters:
+* `repositoryUrl` - a GitHub HTTPS URL.
+* `releases` - a list of comma-separated release(tag) names.
+* `modules` - a list of comma-separated module names.
+
+An example is provided below:
+```Bash
+./generate.sh repositoryUrl='https://github.com/SpineEventEngine/core-java.git' releases='v1.8.0,v1.7.0' modules='core,client'
+```
+
```
armiol, line 41: We need to describe the result which is expected to be outcome of this particular example command.

scripts/dokka/generate.sh
```diff
@@ -0,0 +1,122 @@
...
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+USAGE_TEMPLATE="Usage: generate.sh repositoryUrl='' releases='x,y,z' modules='x,y,z'"
+USAGE_EXAMPLE="Example: generate.sh repositoryUrl='https://github.com/SpineEventEngine/core-java.git' releases='v1.8.0,v1.7.0' modules='core,client'"
+
+#Check that exactly three parameters were provided.
```
armiol, line 32: In all cases, please allow a space symbol after `#`.

scripts/dokka/generate.sh
```diff
@@ -0,0 +1,122 @@
...
+    rm -rf "../$module"
+    log "Finished working on the $module module for the $release release"
+  done
+
+  git push
+
+  rm "../version.gradle.kts"
+  rm "../settings.gradle.kts"
+
+  log "Finished working on the $release release"
```
armiol, line 117: And complete all the sentences in the log messages with periods.

armiol (changes requested):
@tripolkaandrey LGTM in general. Please see my comments though.

scripts/dokka/README.md
```diff
@@ -0,0 +1,150 @@
...
+Description of parameters:
+* `repositoryUrl` - a GitHub HTTPS URL.
+* `releases` - a list of comma-separated release(tag) names. The release with an asterisk will 
+overwrite the published 'primary' one. The release without an asterisk is considered 'secondary', 
+so it is published only to the 'v' directory.
+* `modules` - a list of comma-separated module names.
+
+An example is provided below:
+```Bash
+./generate.sh repositoryUrl='https://github.com/SpineEventEngine/core-java.git' releases='v1.7.0,v1.8.0*' modules='core,client'
```
armiol, line 41: Can we please move the asterisk to the beginning? I.e. `*v1.8.0`. This would allow us to avoid any confusion with masking, as "1.8.0*` could mean "all versions starting from `1.8.0`".

Maybe even some other symbol would fit better.

scripts/dokka/README.md
```diff
@@ -0,0 +1,150 @@
...
+### Usage
+
+The script should be launched from this directory and follow the template provided below:
+```Bash
+./generate.sh repositoryUrl='' releases='x,y,z*' modules='x,y,z'
+```
+
+Description of parameters:
+* `repositoryUrl` - a GitHub HTTPS URL.
+* `releases` - a list of comma-separated release(tag) names. The release with an asterisk will
```
armiol, line 34: We need to be more precise here. Is it a release or a tag? I would say we rely on the fact that each release is accompanied with the corresponding tag. But this script looks for tags only.

scripts/dokka/README.md
```diff
@@ -0,0 +1,150 @@
+### Description
+
+This directory contains a script and serves as a workspace for it. The script generates and publishes
+Dokka documentation for 'old'(`<2.0.0`) releases of Spine projects. Documentation is published to 
+the `gh-pages` branch of the target repository.
+
```
armiol, line 6: Do we clean the "workspace" after the publishing is done?

scripts/dokka/README.md
```diff
@@ -0,0 +1,150 @@
+### Description
```
armiol, line 1: I would rename the folder into something like `scripts/publish`. Here, we'd want to emphasise the functionality, not the tool.

scripts/dokka/generate.sh
```diff
@@ -0,0 +1,142 @@
...
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+PRIMARY_MARK="*"
+USAGE_TEMPLATE="Usage: generate.sh repositoryUrl='' releases='x,y$PRIMARY_MARK,z' modules='x,y,z'"
+USAGE_EXAMPLE="Example: generate.sh repositoryUrl='https://github.com/SpineEventEngine/core-java.git' releases='v1.8.0$PRIMARY_MARK,v1.7.0' modules='core,client'"
```
armiol, line 30: Again, are these releases or tags? It's important, as we might have some discrepancy between the name of the release and its tag (due to a typo, for instance).

armiol (changes requested):
@tripolkaandrey LGTM in general. However, please see my comments.

scripts/publish-documentation/README.md
```diff
@@ -0,0 +1,152 @@
+### Description
+
+This directory contains a script and serves as a workspace for it. The script generates and publishes 
+Dokka documentation for 'old'(`<2.0.0`) releases of Spine projects. Documentation is published to the
```
armiol, line 4: I would say "previous" or "already published" instead of "old". They are hardly old. 1.8.x is the most recent one.

scripts/publish-documentation/buildSrc
```diff
@@ -0,0 +1 @@
+../../buildSrc
```
armiol, line 1: Looks weird to me. But if it works, then fine.

scripts/publish-documentation/publish.sh
```diff
@@ -0,0 +1,139 @@
...
+do
+  if [[ $tag == "$PRIMARY_MARK"* ]]; then
+    # Remove the leading $PRIMARY_MARK.
+    tag=${tag:1}
+    is_primary=true
+  else
+    is_primary=false
+  fi
+
+  log "Started working on the $tag tag."
```
armiol, line 74: Could you please enclose the variable values into backticks? It's here and below. E.g. 
```
log "Started working on the `$tag` tag."
```
I am not sure if it works this way. But in any case, some formatting would feel nice in such cases.

armiol (approved):
@tripolkaandrey LGTM with some minor comments to address.

PR #381, opened 2022-06-15
Support nested modules when publishing Dokka documentation

tripolkaandrey:
### Main changes

This PR adds support for nested modules in the script for publishing Dokka documentation. This is done by changing the API  of the script, so it operates with paths to modules instead of module names. 

### Additional changes

Also, this PR fixes a minor flaw with creating the `gh-pages` if it does not exist. The problem was with the script looking only through local branches which leads to the branch being created even if a remote one already exists.


armiol (approved):
LGTM.

Issue #383, opened 2022-07-07
Use idiomatic Kotlin preconditions in `io.spine.internal.gradle.git.Config`


alexander-yevsyukov:
Now the `Config` class uses precondition checks from `com.google.api.client.util.Preconditions`. This is suspicions even if it were a Java class because it adds dependency on Google API client library rather than on Guava.

Being a Kotlin class it should use idiomatic preconditions. So, instead of

```kotlin
Preconditions.checkArgument(
          name.isNotBlank(),
          "Name cannot be an empty string."
)
```
it should be:

```kotlin
check(name.isNotBlank()} { "Name cannot be an empty string." }
```

Also, please think of a better name for the class. `Config` is way too generic. Assuming it is about the user name and his email, it can be `UserInfo` or `UserAccount` or something like this.

alexander-yevsyukov:
Please also adjust the class documentation. It's not `Wrapper for different Git settings.`, at least in its current form.

PR #384, opened 2022-07-08
Use idiomatic Kotlin preconditions

tripolkaandrey:
This PR addresses #383. 

The main issue was with using `com.google.api.client.util.Preconditions.checkArgument` in `io.spine.internal.gradle.git` package. Firstly, there is a built-in Kotlin function for doing the same thing. Secondly, this `checkArgument` depends on the Google API client library rather than Guava. This PR fixes this.

Also, there was a naming issue with `io.spine.internal.gradle.git.Config`. This PR removes the excessive nesting and leaves only the setting needed for our current needs.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/gradle/git/Repository.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/gradle/git/Repository.kt
@@ -60,11 +60,12 @@ class Repository : AutoCloseable {
      *
      * This configuration determines what ends up in author and commiter fields of a commit.
      */
-    var user: Config.User
+    var user: UserInfo
         get() = field
         private set(value) {
             field = value
         }
+
     /**
      * Currently checked out branch.
      */
@@ -75,7 +76,7 @@ class Repository : AutoCloseable {
         }
 
 
-    private constructor(sshUrl: String, user: Config.User, branch: String) {
+    private constructor(sshUrl: String, user: UserInfo, branch: String) {
         this.sshUrl = sshUrl
         this.user = user
         this.currentBranch = branch
@@ -110,7 +111,7 @@ class Repository : AutoCloseable {
      * values from [user]. These settings determine what ends up in author and
      * commiter fields of a commit.
      */
-    fun configureUser(user: Config.User) {
+    fun configureUser(user: UserInfo) {
         repoExecute("git", "config", "user.name", user.name)
         repoExecute("git", "config", "user.email", user.email)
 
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/kotlin/io/spine/internal/gradle/git/UserInfo.kt
```diff
@@ -26,37 +26,24 @@
...
-                    name.isNotBlank(),
-                    "Name cannot be an empty string."
-                )
-                Preconditions.checkArgument(
-                    email.isNotBlank(),
-                    "Email cannot be an empty string."
-                )
+data class UserInfo private constructor(val name: String, val email: String) {
+    companion object Factory {
+        /**
```
armiol, line 36: Let's add an empty line above.

armiol (approved):
@tripolkaandrey LGTM with one more minor comment.

buildSrc/src/main/kotlin/io/spine/internal/gradle/git/Repository.kt
```diff
@@ -159,17 +159,14 @@ class Repository : AutoCloseable {
          *
          * @throws IllegalArgumentException if SSH URL is an empty string.
          */
-        fun of(sshUrl: String, user: Config.User, branch: String = Branch.master): Repository {
-            Preconditions.checkArgument(
-                !sshUrl.isBlank(),
-                "SSH URL cannot be an empty string."
-            )
+        fun of(sshUrl: String, user: UserInfo, branch: String = Branch.master): Repository {
```
alexander-yevsyukov, line 162: @tripolkaandrey, if it comes to such a static factory method (or even several similar ones), having a named factory object in Kotlin is too much of a ceremony. 

In Java we would write `UserInfo.of(...)` using a static factory method, and it feels natural. Why have `UserInfo.Factory.of(...)` in Kotlin? It's longer to read and longer to type.

A named `Factory` class (rather than object) would be appropriate when we have some parameters passed to it. Or, we have the state of the factory, so it would start producing slightly different objects.

Otherwise, static methods are just fine.

buildSrc/src/main/kotlin/io/spine/internal/gradle/git/UserInfo.kt
```diff
@@ -26,37 +26,25 @@
...
-                    "Email cannot be an empty string."
-                )
+    companion object Factory {
+        /**
+         * Validates provided parameters and constructs a [UserInfo] object.
+         *
+         * @throws IllegalArgumentException if the name or the email is an empty
+         *         string.
+         */
+        fun of(name: String, email: String): UserInfo {
```
alexander-yevsyukov, line 43: @tripolkaandrey, and why not have just a public constructor for `UserInfo`? 

We don't cache these objects, we don't do anything  special about them. If it's just for checking preconditions, then `init {}` section in Kotlin serves this for this purpose just fine.

buildSrc/src/main/kotlin/io/spine/internal/gradle/git/Repository.kt
tripolkaandrey: I used `.Factory.` explicitly because it avoids hiding the type's documentation. Otherwise, when you have `UserInfo.of(...)` and you hover over the `UserInfo` part there is documentation about the `UserInfo.Factory` instead of the `UserInfo`.

PR #385, opened 2022-07-11
Fix issues with `io.spine.internal.gradle.git`

tripolkaandrey:
This PR continues fixing issues with `io.spine.internal.gradle.git`. More specifically, it addresses comments of @alexander-yevsyukov in the #384 after it was merged. 

The first issue was using factory methods the following way `ClassName.Factory.of`. At the time of development, IDEA did not allow traversing from a companion object's documentation to the enclosing class' documentation. It seemed reasonable to state the `Factory` part explicitly. Currently, there is no such issue, so this part is removed.

The second issue was with the `UserInfo` factory method. This factory method, at this point, can be reduced to a simple validation in the `init{}` block.


alexander-yevsyukov (approved):
LGTM

armiol (approved):

PR #391, opened 2022-09-26
Document `desc.ref` file usage

alexander-yevsyukov:
This PR adds documentation on the creation of `desc.ref` files in the modules that manually configure `GenerateProtoTask` via `setup()` extension task provided by `ProtoTaskExtensions.kt`.


buildSrc/src/main/kotlin/io/spine/internal/gradle/protobuf/ProtoTaskExtensions.kt
```diff
@@ -60,6 +61,16 @@ import org.gradle.kotlin.dsl.get
  * Using the same code under `subprojects` in a root build file does not seem to work because
  * test descriptor set files are not copied to resources. Performing this configuration from
  * subprojects solves the issue.
+ *
+ * IMPORTANT: In addition to calling `setup`, a submodule must contain d descriptor set reference
```
dmdashenkov, line 65:
```suggestion
 * IMPORTANT: In addition to calling `setup`, a submodule must contain a descriptor set reference
```

dmdashenkov (approved):

PR #392, opened 2022-09-27
Use action versions already used by subprojects

alexander-yevsyukov:
This PR updates GitHub workflows with action versions already used by some of the subprojects.


armiol (approved):
LGTM.

PR #398, opened 2022-10-12
Add `Spine` dependency object

alexander-yevsyukov:
This PR adds the dependency object `Spine` with the aim to standardize internal project dependencies across all repositories in the near future. 

The implementation was taken from `mc-java` and was slightly refactored for brevity.


alexander-yevsyukov:
Please note that using this object would require renaming some current project extra properties.

For example, it should be `baseVersion` instead `spineBaseVersion`, as some projects still use.

dmdashenkov (approved):

PR #400, opened 2022-10-13
Fix Kotlin dependency suppression

alexander-yevsyukov:
This PR brings modifications to `ProtoTaskExtensions.kt` [introduced in `tool-base`](https://github.com/SpineEventEngine/tool-base/pull/28/files#diff-c9e279e722d4a87b541fe8ef7efd4d022c9fa4f87e2375cb70f12581116139a) adding the following improvements:
  * Misspelling in the file name comment for "kotlin" was fixed.
  * `"kotlin"` built-in for `protoc` is now added conditionally to handle the case when it was already added by the calling build code.
  * Documentation updated with the new steps.
  * Kotlin warnings were addressed.


buildSrc/src/main/kotlin/io/spine/internal/gradle/protobuf/ProtoTaskExtensions.kt
```diff
@@ -121,10 +112,84 @@ fun GenerateProtoTask.setup(generatedDir: String) {
...
+ * classes that duplicate those coming from Protobuf library JARs.
+ */
+private fun Task.deleteComGoogle(generatedDir: String, ssn: String, language: String) {
+    val comPackage = File("${generatedDir}/${ssn}/$language/com")
+    val googlePackage = comPackage.resolve("google")
+
+    project.delete(googlePackage)
+
+    // We don't need an empty `com` package.
+    if (comPackage.exists() && comPackage.list()?.isEmpty() == true) {
```
dmdashenkov, line 133: I'd add an extra condition to check if it's a directory and suppress nullability with `!!` after. Seems more descriptive that way.

dmdashenkov (approved):
LGTM with a suggestion.

PR #407, opened 2022-10-21
Add `change` to `Spine` dependency object

alexander-yevsyukov:
This PR extends the `Spine` dependency object with recently introduced `change` module.

The spellcheck dictionary of IntelliJ IDEA was extended with words recently added in `core-java`.

Also ErrorProne was bumped to `2.16` (which was tested under `core-java`).

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/dependency/Spine.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/dependency/Spine.kt
@@ -82,6 +82,12 @@ class Spine(p: ExtensionAware) {
         const val time = "2.0.0-SNAPSHOT.109"
 
         /**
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/io/spine/internal/dependency/Spine.kt
```diff
@@ -81,6 +81,12 @@ class Spine(p: ExtensionAware) {
          */
         const val time = "2.0.0-SNAPSHOT.109"
 
+        /**
+         * Tghe version of `change` to use.
```
armiol, line 85: Please fix the typo.

armiol (approved):
LGTM with a comment to address.

PR #422, opened 2022-10-28
Perform Detekt code analysis with a Gradle task

yevhenii-nadtochii:
This PR sets up Detekt code analysis with a Gradle task.

Usually, it's done during the build, but `config` is not building at all. GitHub workflow is still used, but now it runs `detekt` task. This workflow is a special case for `config` repository. `pull` will not copy it along with other workflows to destination repos.

Detekt's findings I've marked as known to address them in the following PR separately.

Side changes:

1. `config-tester` script-plugin is extracted from root's `build.gradle.kts`. It makes the root script more declarative.
2. `detekt-code-analysis` plugin is configured to generate only HTML reports. We don't need all five types of reports.
3. Top-level docs to `detekt-code-analysis` plugin is attached to `about` variable to make them render.


yevhenii-nadtochii:
@armiol @alexander-yevsyukov Please take a look.

alexander-yevsyukov (approved):
LGTM

Issue #214, opened 2021-07-05
Forced JaCoCo version


dmdashenkov:
There is an issue with JaCoCo's parser which does not allow it to work with some classes generated from Kotlin. The issue has been fixed in version `0.8.7`. However, Gradle, at least as of version `6.9`, has not yet caught up with the JaCoCo update.

For now, we've created the `force-jacoco.gradle.kts` script to work around the issue. If there's a JaCoCo error related to Kotlin sources, try applying this script.

We should keep an eye on Gradle updates and check if the fix has been propagated from time to time.

PR #428, opened 2022-10-31
Address Detekt findings

yevhenii-nadtochii:
This PR addresses numerous findings reported by Detekt for code in `buildSrc`.

Most of them were addressed in code. Several cases are suppressed. 

`force-jacoco` script-plugin was dropped. The [reason](https://github.com/SpineEventEngine/config/issues/214) of its existence seems to be resolved in Gradle `7.1.-rc-1`.

From [Gradle 7.1-rc-1 Release Notes](https://docs.gradle.org/7.1-rc-1/release-notes.html):

>[The JaCoCo plugin](https://docs.gradle.org/7.1-rc-1/userguide/jacoco_plugin.html) has been upgraded to the most recent [JaCoCo version 0.8.7](http://www.jacoco.org/jacoco/trunk/doc/changes.html) which includes support for Java 15 and 16 and experimental support for Java 17.

I've tried to build `base` without this plugin, and it builds fine. Our current version of Gradle fetches JaCoCo `0.8.8`. 

`Scripts.kt` referenced scripts which don't exist anymore and was also deleted.

Detekt configuration was updated as the following:

1) `UnusedPrivateMember` now passes variables named `about`.
2) `MagicNumber` was updated to include `3`. 
3) `MaxLineLength` was updated to ignore comments due to [long links issue](https://github.com/detekt/detekt/issues/5493).


yevhenii-nadtochii:

alexander-yevsyukov:
> UnusedPrivateMember now passes variables named about

Can we please have `ABOUT` included too (per my recent comment were you were at-mentioned)?

buildSrc/src/main/kotlin/force-jacoco.gradle.kts
```diff
@@ -1,39 +0,0 @@
...
-    configurations.all {
-        resolutionStrategy {
-            eachDependency {
-                if (requested.group == "org.jacoco") {
-                    useVersion("0.8.7")
-                }
-            }
-        }
-    }
-}
```
alexander-yevsyukov, line 39: Please update the `pull` script so that it removes this file when installing the latest version.

buildSrc/src/main/kotlin/io/spine/internal/gradle/ConfigTester.kt
```diff
@@ -302,31 +311,31 @@ class ClonedRepo(
...
-            }
+            copyRecursively(sourceFolder, ignoredFolder, destinationFolder)
         } catch (e: Exception) {
             throw IllegalStateException(
                 "Error copying folder `$sourceFolder` to `$destinationFolder`.", e
             )
         }
     }
+
+    private fun copyRecursively(sourceFolder: Path, ignoredFolder: Path?, destinationFolder: Path) {
```
alexander-yevsyukov, line 322: Don't we have have it natively available from Kotlin?

buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
```diff
@@ -54,25 +54,20 @@ data class Repository(
      * the credentials. The file must have properties `user.name` and `user.password`, which store
      * the username and the password for the Maven repository auth.
      */
-    fun credentials(project: Project): Credentials? {
+    fun credentials(project: Project): Credentials? =
         if (credentialValues != null) {
-            return credentialValues.invoke(project)
+            credentialValues.invoke(project)
+        } else {
+            val log = project.logger
```
alexander-yevsyukov, line 61: Please extract lines until `creds` into a method.

buildSrc/src/main/kotlin/io/spine/internal/gradle/publish/IncrementGuard.kt
```diff
@@ -80,13 +80,22 @@ class IncrementGuard : Plugin<Project> {
...
             return false
         }
-        val reference = System.getenv("GITHUB_REF") ?: return false
-        val matches = Regex("refs/heads/(.+)").matchEntire(reference) ?: return false
-        val branch = matches.groupValues[1]
-        return !branch.endsWith("master")
+        val branch = branchName(reference)
+        return when {
+            branch == null -> false
+            branch.endsWith("master") -> false
```
alexander-yevsyukov, line 91: We need more options for the main branch, just in case we'd have to default to `"main"`.

I don't want to do that because I don't like these attempts to impose "newspeak". But we may be _forced_ to change the master branch name. So, let's be ready.

buildSrc/src/main/kotlin/force-jacoco.gradle.kts
alexander-yevsyukov: Please also take care for removal of `deps-between-tasks.kt` because it introduces conflicts with `Dependencies.kt` (to which it was renamed). But it takes manual operation to remove the file once compiler complains about it. And it's not usually obvious. So, let's same some time for our colleagues.

alexander-yevsyukov: Please also don't forget to close the referenced issue.

alexander-yevsyukov (changes requested):
Please see my questions and comments.

armiol (approved):
LGTM from my side, apart from the points raised by Alexander. 

Approving now, so that I don't block the PR even more.

buildSrc/src/main/kotlin/force-jacoco.gradle.kts
yevhenii-nadtochii: Can we just delete the whole destination `buildSrc` and then copy it from scratch? Is this safe for all repos? In other words, do we have repos that have custom code in their `buildSrc`?

I'm asking because this clean-up looks a bit fragile:

```bash
echo "Updating Gradle 'buildSrc' scripts"
rm -f ../buildSrc/src/main/kotlin/force-jacoco.gradle.kts
rm -f ../buildSrc/src/main/kotlin/deps-between-tasks.kt
rm -f ../buildSrc/src/main/kotlin/io/spine/internal/gradle/Scripts.kt
cp -R buildSrc ..
```

Now `pull` script always copies "on top" that leads to accumulation of old files.

buildSrc/src/main/kotlin/io/spine/internal/gradle/ConfigTester.kt
yevhenii-nadtochii: Yes, partially. Kotlin has `File.copyRecursively()` but it doesn't allow specifying exclusions. We could copy the whole directory and then delete asked `ignoredFolder` but it would come with some penalty in case `ignoredFolder` is big.

buildSrc/src/main/kotlin/force-jacoco.gradle.kts
alexander-yevsyukov: > Is this safe for all repos? 
I don't think it's a good idea. Too aggressive. People may have custom stuff there. 

Sure, we do have Git, and our colleagues would be able to restore the deleted files. But it's too annoying.
Computers should annoy people less, not more.

alexander-yevsyukov: > Now pull script always copies "on top" that leads to accumulation of old files.

That's why we need to clean up things with each next version. And then remove the cleanup code, after some time, too.

yevhenii-nadtochii:
@alexander-yevsyukov Please take another look.

alexander-yevsyukov (approved):
LGTM

PR #436, opened 2022-11-04
Escape our dependency objects with plugin accessors

yevhenii-nadtochii:
This changeset updates escaped dependency objects to be plugin accessors. 

Reasoning is the following:

1. Those escaped dependencies seems to be used only in `plugins` sections, and the only way to use them there is to apply plugins.
2. Script-plugins also have such accessors generated along with the ones, provided by Gradle. Thus, in many cases, we can drop all `id(<string litaral>)` calls, making `plugins` section look coherent.
3. Custom versions can still be declared.
4. Names of those accessors should match [naming conventions](https://docs.gradle.org/current/userguide/custom_plugins.html#sec:creating_a_plugin_id).

This change was previously adopted in [`base-types` #31](https://github.com/SpineEventEngine/base-types/pull/31).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -38,9 +38,3 @@ detekt {
     source.from("buildSrc/src/main/kotlin")
     config = files("quality/detekt-config.yml")
 }
-
-doctor {
-    javaHome {
-        failOnError.set(false)
-    }
-}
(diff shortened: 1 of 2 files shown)


yevhenii-nadtochii:

armiol (approved):

PR #437, opened 2022-11-04
Drop `protodata` and `mc-java` from plugins accessors and update Detekt config

yevhenii-nadtochii:
This changeset does the following:

1. Moves `protodata` and `mc-java` plugin accessors to the corresponding dependency accessors: `protoData` and `mcJava`. `mc-java` cannot be applied to a project directly. `protodata` can be, but when used in pair with `mc-java`, buildscript classpath ends up with both of them. Thus, making impossible direct apply of `protodata` too.
2. Updates Detekt config to exclude `*Extensions.kt` files from `TooManyFunctions` rule. Such files usually consists only of  functions.

### DIFF
--- a/buildSrc/src/main/kotlin/PluginsAccessors.kt
+++ b/buildSrc/src/main/kotlin/PluginsAccessors.kt
@@ -30,8 +30,6 @@
 import io.spine.internal.dependency.ErrorProne
 import io.spine.internal.dependency.GradleDoctor
 import io.spine.internal.dependency.Protobuf
-import io.spine.internal.dependency.Spine
-import io.spine.internal.dependency.Spine.ProtoData
 import org.gradle.plugin.use.PluginDependenciesSpec
 import org.gradle.plugin.use.PluginDependencySpec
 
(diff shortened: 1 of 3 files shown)


yevhenii-nadtochii:

armiol (approved):
@yevhenii-nadtochii LGTM, however, I would hear from @alexander-yevsyukov  on this matter, as he also works closely with both ProtoData and model compiler at the moment.

buildSrc/src/main/kotlin/PluginsAccessors.kt
```diff
@@ -69,6 +64,3 @@ val PluginDependenciesSpec.protobuf: PluginDependencySpec
 
 val PluginDependenciesSpec.`gradle-doctor`: PluginDependencySpec
     get() = id(GradleDoctor.pluginId).version(GradleDoctor.version)
-
-val PluginDependenciesSpec.`mc-java`: PluginDependencySpec
-    get() = id(Spine.McJava.pluginId).version(Spine.McJava.version)
```
alexander-yevsyukov, line 74: I get the point. This is rough. But it makes sense.

alexander-yevsyukov (approved):
LGTM

PR #439, opened 2022-11-07
Make build workflow fail if no tests were executed

yevhenii-nadtochii:


yevhenii-nadtochii:

pull
```diff
@@ -89,6 +89,8 @@ echo "Cleaning up Gradle 'buildSrc' scripts"
 rm -f ../buildSrc/src/main/kotlin/force-jacoco.gradle.kts
 rm -f ../buildSrc/src/main/kotlin/deps-between-tasks.kt
 rm -f ../buildSrc/src/main/kotlin/io/spine/internal/gradle/Scripts.kt
+rm -f ../buildSrc/src/main/kotlin/io/spine/internal/gradle/test/Logging.kt
+rm -f ../buildSrc/src/main/kotlin/io/spine/internal/gradle/test/Tasks.kt
```
armiol, line 93: It's a bit fragile. Can we clean everything that's in these folders?

I don't think it's too convenient to remember this place has to be updated for each new file.

armiol (approved):
LGTM with a comment to take into account.

pull
yevhenii-nadtochii: We can clean the whole `io/spine/internal/gradle/test` folder. But it doesn't work for other folders here. They contain relevant files as well.

PR #441, opened 2022-11-08
Force Dokka library

yevhenii-nadtochii:
This PR forces the version of Dokka base library. 

Dokka is widely used now among our modules and the same lib is usually forced.

By the way, this PR does the following:

1. Bumps `grGit` to `4.1.1`. The currently used version is not unavailable because JCenter is deprecated now. And bumping the version to the one, which is available in Maven Central, seems to be a good alternative to [registering](https://github.com/ajoberstar/grgit#old-versions-from-bintrayjcenter) another repository.
2. Allows long method names in `test` and `spec` sources.
3. Fixes numerous typos.


yevhenii-nadtochii:

PR #443, opened 2022-11-11
Update Detekt configuration

yevhenii-nadtochii:
This PR updates Detekt config to exclude `*View.kt` and `*Projection.kt` files from `TooManyFunctions` rule.

Inheritors of `Projection` usually have a number of subscribing methods, so I propose to exclude them from this rule.

Also, this PR defines an [anchor](https://yaml.org/spec/1.2-old/spec.html#id2785586) for test and spec files to re-use them in filters to other rules.


yevhenii-nadtochii:

armiol (approved):

PR #444, opened 2022-11-15
Rename a workflow

yevhenii-nadtochii:
This PR renames a workflow to be consistent with others, which use kebab case.

`ensure-reports.updated.yml` becomes `ensure-reports-updated.yml`. `pull` script is updated to delete the previous workflow to avoid duplication.


yevhenii-nadtochii:

alexander-yevsyukov (approved):
LGTM

PR #448, opened 2023-01-12
Include `generated-proto` directories in PMD

alexander-yevsyukov:
This PR configures PMD to ignore `generated-proto` directories that are used by ProtoData as an interim location for source code generated by Protobuf Compiler.

Sometimes these directories are picked up by PMD during the build, which fails it because of numerous issues in the generated code. Ideally, `generated-proto` directories should not be seen by PMD as source code directories, but teaching ProtoData to do such tricks as a separate task.

This PR configures PMD do not look into them in any case, so that we can build anyway.


armiol (approved):
LGTM, but let's update the PR name to say "exclude" somewhere. Right now it reads as if we were to add these folders to those checked by PMD.

PR #470, opened 2023-05-25
Adjust Dokka publishing according to the latest Dokka changes

armiol:
This changeset updates Dokka publishing script.

Previously, Dokka output was always residing in `dokka` sub-folder. However, starting from some version of Dokka (and it's not too easy to tell from which exactly), the output for Java code is now located in `dokkaJava`.

This PR addresses this change in Dokka behaviour by modifying `publish.sh` script accordingly.

Other changes:

* More logging was added to the publishing script in order to better track potential failures.
* Samples on using `publish.sh` for `1.x` artifacts were added, so that end-users spend less time searching for proper parameter values.

### DIFF
--- a/scripts/publish-documentation/publish.sh
+++ b/scripts/publish-documentation/publish.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 #
-# Copyright 2022, TeamDev. All rights reserved.
+# Copyright 2023, TeamDev. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


armiol:
@tripolkaandrey PTAL.

scripts/publish-documentation/publish.sh
```diff
@@ -116,22 +119,38 @@ do
...
   for path in $(echo "$paths" | tr "," "\n")
   do
     # Extracts the module name from a path. A path should use the "/" as a file separator.
     module="${path##*\/}"
 
+    log "[$PWD] Creating directory 'dokka-reference/$module/v/$version'"
     mkdir -p "dokka-reference/$module/v/$version"
-    cp -r "../$module/build/docs/dokka/" "dokka-reference/$module/v/$version/"
+
+    # Previous versions of Dokka produced their output in `dokka` subfolder.
```
tripolkaandrey, line 137: I don't think we should document this change in the output directory. This description explains well what this PR is about, but there is no reason why the next engineer who looks at the script after the PR is merged may need the information about how it was before.

scripts/publish-documentation/publish.sh
```diff
@@ -116,22 +119,38 @@ do
...
     if [ $is_primary = true ]; then
-      cp -r "../$module/build/docs/dokka/" "dokka-reference/$module/"
+      log "[$PWD] PRIMARY Copying recursively: '../$module/build/docs/dokkaJava/' -> 'dokka-reference/$module/'"
+      cp -r "../$module/build/docs/dokkaJava/" "dokka-reference/$module/"
       commit_message+="as primary"
     else
       commit_message+="as secondary"
     fi
 
+    log "[$PWD] Executing 'git add \"dokka-reference/$module/*\"', and committing."
```
tripolkaandrey, line 153: `Staging and committing all changes in the \"dokka-reference/$module/*\" `seems shorter and conveys the same meaning.

PR #475, opened 2023-07-11
Update dependencies, simplify script plugins

alexander-yevsyukov:
This PR:
  * Bumps versions of Kover, detekt, and ErrorProne Gradle plugin.
  * Merges `java-module` script plugin into `kotlin-jvm-module.


buildSrc/src/main/kotlin/jacoco-kmm-jvm.gradle.kts
yevhenii-nadtochii, line 1: Should we use KMM abbreviation? It stands for Kotlin Multiplatform Mobile. 

We could go with KM or KMP.

buildSrc/src/main/kotlin/kotlin-jvm-module.gradle.kts
yevhenii-nadtochii, line 1: Or simpler `jvm-module`. And think of Spine JVM module as one that can be Java or Kotlin, or mixed.

alexander-yevsyukov: Let's go with `jvm-module`. I'll rename.

buildSrc/src/main/kotlin/jacoco-kmm-jvm.gradle.kts
alexander-yevsyukov: I've made it simply `jacoco-kotlin-jvm` because it would tell directly what it's for.

yevhenii-nadtochii (approved):

PR #480, opened 2023-07-23
`Logging` dependency object

alexander-yevsyukov:
This PR introduces `Spine.Logging` dependency object to gather multiple artifacts produced by this subproject under its "umbrella".


alexander-yevsyukov:
Let's start moving towards a separate library (while we selecting a good name for it).

yevhenii-nadtochii (approved):

PR #481, opened 2023-07-25
Bump Spine Logging

yevhenii-nadtochii:
This PR bumps the version of `spine-logging` to the latest one, in which the published library is operational again.

Also, it substitutes a runtime dependency on Flogger's JUL backend with Spine Logging. In general, specifying of a backend makes no sense unless a facade is used. I think we can go with the following options:

1. Configure Spine logging by default with our facade and JUL backend. This option is employed in this PR.
2. Doesn't pre-configure logging at all. Remove specifying of Flogger's backend.

We **potentially can** detect which facade is used, and then specify the corresponding backend in `jvm-module`. But this will have to be done with `project.afterEvaluate { ... }` closure. Because if `jvm-module` doesn't dictate the used facade, then a concrete module is meant to dictate it. A script plugin is applied first, it can't obtain module-configured dependencies in advance. Thus, I would rather avoid this sort of tricks that tampers with Gradle configuration phases.

The version of `org.ow2.asm:asm` transitive dependency has been forced to prevent conflicts.

### DIFF
--- a/buildSrc/src/main/kotlin/jvm-module.gradle.kts
+++ b/buildSrc/src/main/kotlin/jvm-module.gradle.kts
@@ -149,6 +149,9 @@ fun Module.addDependencies() = dependencies {
     compileOnlyApi(JavaX.annotations)
     ErrorProne.annotations.forEach { compileOnlyApi(it) }
 
+    implementation(Spine.Logging.lib)
+    runtimeOnly(Spine.Logging.backend)
+
     testImplementation(Guava.testLib)
     testImplementation(JUnit.runner)
     testImplementation(JUnit.pioneer)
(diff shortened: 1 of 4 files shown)


yevhenii-nadtochii:

buildSrc/src/main/kotlin/jvm-module.gradle.kts
```diff
@@ -159,8 +162,6 @@ fun Module.addDependencies() = dependencies {
     testImplementation(Kotest.datatest)
     testImplementation(Kotest.runnerJUnit5Jvm)
     testImplementation(JUnit.runner)
-
-    runtimeOnly(Flogger.Runtime.systemBackend)
```
armiol, line 163: Given the description you provided for the PR, can you please try applying the fresh `config` to any of our recently updated repositories (such as `mc-java`, `validation`, `ProtoData`, `base`, etc), and verify that this is the correct approach and it does not break anything.

armiol (approved):
@yevhenii-nadtochii LGTM with a comment to process.

buildSrc/src/main/kotlin/jvm-module.gradle.kts
yevhenii-nadtochii: Ok, I will verify.

PR #484, opened 2023-07-30
Small fixes

alexander-yevsyukov:
This PR:
  * Bumps gRPC.
  * Bumps internal dependencies.
  * Improves `Runtime.kt` file removing unused code and making it more idiomatic for Kotlin.
  * Makes `ExcludeInternalDoclet` class to use artifact declared in `Spine` object.
  * Extends `pull` script to take care of removal of `.lift.toml` file.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/gradle/Runtime.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/gradle/Runtime.kt
@@ -33,11 +33,6 @@ import java.io.InputStream
 import java.io.StringWriter
 import java.util.*
 
-object Runtime {
-    @Suppress("unused")
-    val flogger = Flogger.Runtime
-}
-
 /**
  * Executor of CLI commands.
  *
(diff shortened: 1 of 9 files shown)


buildSrc/src/main/kotlin/io/spine/internal/gradle/Runtime.kt
```diff
@@ -26,17 +26,17 @@
...
 
-object Runtime {
-    @Suppress("unused")
-    val flogger = Flogger.Runtime
-}
+/**
+ * Utilities for working with processes from Gradle code.
+ */
+@Suppress("unused")
+private const val ABOUT = ""
```
armiol, line 39: Shouldn't we say something here?

armiol (approved):
@alexander-yevsyukov LGTM with a comment to consider.

buildSrc/src/main/kotlin/io/spine/internal/gradle/Runtime.kt
alexander-yevsyukov: The point of having this constant is to have a KDoc, which covers the purpose of the file.  We do it this way in some other files under `buildSrc`.

Having the constant, in turn, also allows to avoid renaming the file. Now, after removing the `Runtime` object from this file, there's only one top-level declaration which is `Cli` class. detekt insists on having the file named after single top-level declaration. That's why I added the constant.

PR #488, opened 2023-09-21
Local ProtoData versions

dmdashenkov:
In this PR we make the `buildSrc`'s script `ProtoData` choose which version of ProtoData to use, a public remote one or one published locally.

In order to switch to the locally published version, declare the `PROTO_DATA_VERSION` env variable (or the `PROTO_DATA_DF_VERSION`) with the desired version. The script will handle the rest.


alexander-yevsyukov (approved):
LGTM

PR #500, opened 2023-11-24
Force Kotlin `stdlib-jdk7` dependency version

armiol:
This PR forces `kotlin-stdlib-jdk7` dependency version.

As it turns out, this dependency comes as conflicting in at least `base`, `reflect` (!), and `logging` libraries. Apparently, some of our direct dependencies brings the old version of this library with it.

After some attempts to find out, which dependency is "guilty", I gave up, as no clear answer is available. So it's easier to force its version.


armiol:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #504, opened 2024-01-02
Improve the generation of descriptor set files for "simple" Spine libraries

armiol:
At the moment, not all Spine libraries are able to use Spine McJava plugin — due to the cyclic dependencies otherwise introduced. But simultaneously, libraries like `tool-base` and `base-types` provide their own Proto language, and hence, must bring respective descriptor set files with them.

This issue has been addressed very inconsistently throughout the repositories, leading to mis-interpretation of descriptor set files, or failures to locate them at all.

This changeset improves the descriptor set generation by doing the following.

* For the descriptor set file name, a Maven coordinates of the respective artifact are used. Previously, we used `known_types_{sourceSetName}.desc` everywhere, which caused clashes.
* Generate a new `desc.ref` file referencing the generated descriptor set file. So far, it was not done at all, leading to loss of the generated descriptor set files.

These changes are inspired by the corresponding code in McJava (see `JavaProtocConfigurationPlugin`). However, in this PR, the code is much simpler, and no extra Gradle tasks is created.


armiol:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #507, opened 2024-02-11
Depend on `buildSrc.jar` in case `src` is missing

armiol:
This changeset equips `buildSrc/build.gradle.kts` with another `implementation`-level dependency, which allows to shorten the build time for integration tests by re-using `buildSrc.jar` instead of re-compiling Kotlin sources in `buildSrc/src`.


alexander-yevsyukov:
@armiol, please consider adding  exclusion of `.gradle-test-kit` to `.gitignore` before merging.

PR #513, opened 2024-05-21
Improve the content of `dependencies.md`, tell others about our license

armiol:
This changeset addresses the issues recently discovered in relation to our own artifacts, and to the list of dependencies we compose.

### `dependencies.md`

Previously, when composing a list of third-party dependencies (a.k.a. `dependencies.md`) we were filtering out Spine artifacts—since we aren't a third party to ourselves. However, as time goes by, more artifacts were introduced. And this changeset filters them out by their group IDs, as well.

It was also found that `dependencies.md` contained a list of artifact versions, some of which were NOT used in real life, because someone could be forcing another version instead. Our previous utilities ignored that fact, which also caused false-alarm reports on using several versions of the same dependency.

This changeset addresses the issue by taking the forced versions into account.

### `pom.xml` and our license

We generate `pom.xml` for each Maven artifact we publish. However, previously we were not mentioning the software version under which Spine is distributed. Now, we explicitly include this content into `pom.xml`:

```xml
...
  <licenses>
    <license>
      <name>The Apache Software License, Version 2.0</name>
      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
    </license>
  </licenses>
...

```

### `MANIFEST.mf` and our license, again

Similarly to `pom.xml`, starting now the `MANIFEST.mf` files also bring the mention of the license, under which a particular JAR is distributed. Previously, there was no such content. Here is an example:

```
<Excerpt from MANIFEST.mf>
Manifest-Version: 1.0
...
Implementation-Title: io.spine.protodata:protodata-cli
Implementation-Version: 0.24.0
Implementation-Vendor: TeamDev
Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
```

:warning: Not all Spine artifacts customise writing of their `MANIFEST.mf`. So the changes made apply only to those Gradle modules, which apply our `write-manifest` plugin-script.

The measures taken will now allow third-party license detectors, such as [Gradle-License-Report] plugin fetch the license for Spine artifacts.

### Version updates

Lastly, Gradle-License-Report plugin was updated to 2.7, its latest release. 
It is already noted they have improved the license detection for some third-party dependencies, so we expect less "license unknown" entries in our dependency reports.

[Gradle-License-Report]:https://github.com/jk1/Gradle-License-Report


alexander-yevsyukov (approved):
LGTM

PR #540, opened 2025-02-26
Adjust the Gradle config for `generatePom` task

armiol:
This PR addresses [base#855](https://github.com/SpineEventEngine/base/issues/855).

This changeset moves the declaration of the dependency between `generatePom` and `build` out of the lazy `Action` and into the explicitly called function. 

Starting Gradle 8.13, the lambda-`Action` of `tasks.register("foo") { ... }` block seems to not be executed like previously, without such an explicit dependency declared. Therefore, in its previous form, in Gradle 8.13 there were no `generatePom` task created.

This change was also tested with Gradle 7.6.4 in `core-java`.


armiol:
@alexander-yevsyukov FYI.

yevhenii-nadtochii (approved):

PR #551, opened 2025-03-26
Bump local dependencies

alexander-yevsyukov:
This PR bumps local dependencies recently adopted in CoreJava so that they are used in other subprojects.


armiol (approved):
@alexander-yevsyukov as discussed, let's use `base@....SNAPSHOT-306`.

@yevhenii-nadtochii FYI, with `306` this PR depends on the publishing of your PR to `base`.

PR #558, opened 2025-04-16
Rollback gRPC Kotlin and force line separators to Unix

alexander-yevsyukov:
This PR rolls back the version of gRPC for Kotlin to 1.4.1 becuse 1.4.2 is not available from public Maven repositories.

Also this PR forces the line separators in the settings because pulling `config` recently gives many warnings about encountered Windows line separators. Let's stick to Unix separators to avoid the noise.

### Changes to `buildSrc`
 * `JUnit` and `Jackson` dependency objects were updated assuming we are going to force them as platforms.
 * `DependencyResolution` was updated accordingly to the previous item.
 * `module-testing` conventions plugin was added for being used for configuring testing in a JVM module.

### `.gitignore` updates
 * Ignore `**/.kotlin/` directories instead of only files under them.
 * Ignore generated Dart code. This is brought from `base`.
 * Remove redundant entries in `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,7 @@
 .fleet/
 
 # Kotlin temp directories.
-**/.kotlin/**
+**/.kotlin/
 
 # IntelliJ IDEA modules and interim config files.
 *.iml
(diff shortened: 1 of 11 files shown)


armiol:
@alexander-yevsyukov Detekt is still failing...

armiol (approved):
@alexander-yevsyukov LGTM with the respect to fixing the build.

PR #574, opened 2025-05-17
Updates from ProtoData

alexander-yevsyukov:
This PR brings changes in `buildSrc` recently adopted in ProtoData. Also, it adds the flag in `gradle.properties` turning on parallel build task execution.


gradle.properties
```diff
@@ -1,3 +1,5 @@
+org.gradle.parallel=true
```
armiol, line 1: Yes, but why? :)

armiol (approved):

Issue #691, opened 2026-06-15
Build-logic audit follow-ups vendored from config (2026-06 compiler audit)


alexander-yevsyukov:
## Context

The 2026-06 technical audit of `SpineEventEngine/compiler` and
`SpineEventEngine/core-jvm-compiler` surfaced several findings that live in
build-logic files **vendored from this repo** (`config`) into every consumer
via the git submodule + `pull` sync: `buildSrc/**`, `.codecov.yml`, and the
workflow templates under `.github-workflows/`. Those files are byte-identical
in the consumer repos and **cannot be durably fixed there** — they must be
fixed here and re-synced.

This issue collects the `[config]`-scoped follow-ups so they are addressed at
the source. They were previously tracked only in the consumer repos'
improvement plans (`.agents/tasks/improvement-plan.md` in `compiler`, and the
equivalent in `core-jvm-compiler`), which now point here. All paths below are
`config`-repo paths verified at filing time; a single fix here resolves the
finding in both consumer repos.

### 1. (S1, Medium → High if scope broadens) Remove the scrambled GitHub PAT from source

`buildSrc/src/main/kotlin/io/spine/gradle/repo/Repositories.kt:58-80` —
`private object Pat` holds a scrambled `read:packages` token, de-scrambled at
runtime into credentials for `https://maven.pkg.github.com/SpineEventEngine/*`.
The KDoc states the scrambling exists *"because GitHub analyzes commits for the
presence of tokens and invalidates them"* — i.e. it deliberately evades GitHub
secret scanning.

Why it matters: the token is trivially extractable by anyone with repo read
access; it is a shared org-wide credential with no rotation story; and
intentionally defeating secret scanning fails third-party security review.

**Blocked on a human decision** (audit open question 1): the actual scope and
owner of the token, and whether GitHub Packages can be dropped as a resolution
source (the Cloud Artifact Registry `europe-maven.pkg.dev` already serves
releases) or instead fed via `GITHUB_TOKEN` / an environment variable. Target
state: a reviewer concludes "no secrets in source" by `grep`, not by trusting a
comment about token scope.

### 2. (Q3, Low) Document or fix the silently-swallowed exception in the npm-audit task

`buildSrc/src/main/kotlin/io/spine/gradle/javascript/task/Check.kt:143` —
`npm("audit")` is wrapped as
`try { … } catch (ignored: Exception) { npm("audit", "--registry", "https://registry.npmjs.eu") }`.
[...]

PR #693, opened 2026-06-15
Handle `null` versions in dependency reports; fix "which/that" doc grammar

alexander-yevsyukov:
## Summary

Two related build-logic changes to `config`'s `buildSrc` (both float to all
consumer repos via the submodule + `pull`):

1. **Handle `null` versions gracefully in the POM dependency report** — a
   functional bug fix.
2. **Fix "which/that" grammar in build-logic KDoc/comments** — a documentation
   pass.

## 1. Handle `null` versions in the POM dependency report

The POM dependency-report generator emitted BOM-managed coordinates (declared
without an explicit version) as the literal `<version>null</version>` in every
consumer repo's generated `docs/dependencies/pom.xml`. `null` is not a valid
Maven version, making the report hard to consume by tooling.

**Root cause.** `DependencyWriter.writeXmlTo` wrote
`"version" { xml.text(dependency.version) }` unconditionally; for a version-less
dependency `dependency.version` is `null`, and the `MarkupBuilder.text` helper
does `value.toString()`, turning `null` into the string `"null"`. The data layer
already knew a null version was possible (`deduplicate()` guards it with
`it.version ?: ""`) — only this emission site did not.

**Fix.** Emit `<version>` only when present, mirroring the existing conditional
for the optional `<scope>` element:

```kotlin
dependency.version?.let { version ->
    "version" { xml.text(version) }
}
```

A version-less dependency now omits `<version>` entirely — the correct Maven
representation for a dependency-managed version — instead of emitting invalid
XML. Covered by a new `DependencyWriterSpec` regression test
(`omit the version of a dependency that declares none`).

Surfaced by a Copilot review on
[SpineEventEngine/compiler#69](https://github.com/SpineEventEngine/compiler/pull/69#discussion_r3415608846).
Pre-existing on `master`; not introduced by that PR. Same family as the vendored
`buildSrc` audit follow-ups tracked in
[config#691](https://github.com/SpineEventEngine/config/issues/691) (fixed
directly here, since it is self-contained).

## 2. Fix "which/that" grammar in build-logic docs

An English wording pass across ~32 `buildSrc/**/*.kt` files, correcting
restrictive relative clauses ("which" → "that") in KDoc and inline comments,
plus an incidental copyright-year bump (2025 → 2026) on the touched files and
two trivial whitespace cleanups. Doc-only: no executable code changed in these
files.

🙌🏻 Updated by @alexander-yevsyukov

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/gradle/publish/GitHubPackages.kt
+++ b/buildSrc/src/main/kotlin/io/spine/gradle/publish/GitHubPackages.kt
@@ -82,8 +82,8 @@ private fun Project.readGitHubToken(): String {
 }
 
 /**
- * Read the personal access token for the `developers@spine.io` account that
- * has only the permission to read public GitHub packages.
+ * Reads the personal access token for the `developers@spine.io` account.
+ * The token grants only read access to public GitHub packages.
  *
  * The token is extracted from the archive called `aus.weis` stored under `buildSrc`.
  * The archive has such an unusual name to avoid scanning for tokens placed in repositories
--- a/buildSrc/src/main/kotlin/io/spine/gradle/publish/GitHubPackages.kt
+++ b/buildSrc/src/main/kotlin/io/spine/gradle/publish/GitHubPackages.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025, TeamDev. All rights reserved.
+ * Copyright 2026, TeamDev. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -82,12 +82,12 @@ private fun Project.readGitHubToken(): String {
 }
 
 /**
- * Read the personal access token for the `developers@spine.io` account which
+ * Read the personal access token for the `developers@spine.io` account that
  * has only the permission to read public GitHub packages.
  *
  * The token is extracted from the archive called `aus.weis` stored under `buildSrc`.
  * The archive has such an unusual name to avoid scanning for tokens placed in repositories
- * which is performed by GitHub. Since we do not violate any security, it is OK to
+ * that is performed by GitHub. Since we do not violate any security, it is OK to
  * use such a workaround.
  */
 private fun Project.readTokenFromArchive(): String {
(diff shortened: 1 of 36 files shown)


buildSrc/src/main/kotlin/io/spine/gradle/publish/GitHubPackages.kt
```diff
@@ -82,12 +82,12 @@ private fun Project.readGitHubToken(): String {
...
 /**
- * Read the personal access token for the `developers@spine.io` account which
+ * Read the personal access token for the `developers@spine.io` account that
  * has only the permission to read public GitHub packages.
  *
  * The token is extracted from the archive called `aus.weis` stored under `buildSrc`.
  * The archive has such an unusual name to avoid scanning for tokens placed in repositories
- * which is performed by GitHub. Since we do not violate any security, it is OK to
+ * that is performed by GitHub. Since we do not violate any security, it is OK to
  * use such a workaround.
```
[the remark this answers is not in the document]
alexander-yevsyukov: This is a part of work on #691, not this PR.

buildSrc/src/main/kotlin/io/spine/gradle/report/pom/DependencyWriter.kt
```diff
@@ -88,7 +88,12 @@ private constructor(
                     "dependency" {
                         "groupId" { xml.text(dependency.group) }
                         "artifactId" { xml.text(dependency.name) }
-                        "version" { xml.text(dependency.version) }
+                        // A BOM-managed dependency carries no explicit version.
+                        // Omit the element rather than emit `<version>null</version>`,
+                        // which is not a valid Maven version.
```
armiol: @claude shouldn't it be "that is not a valid Maven version"?

armiol (changes requested):
@claude see my comments.

buildSrc/src/main/kotlin/io/spine/gradle/report/pom/DependencyWriter.kt
alexander-yevsyukov: I has been re-written with "since".

armiol (approved):

PR #702, opened 2026-06-19
Limit init-submodules bootstrap to config-managed submodules

alexander-yevsyukov:
## Problem

Addresses [#697 (comment)](https://github.com/SpineEventEngine/config/pull/697#discussion_r3432637633).

`init-submodules` runs automatically from a `SessionStart` hook. The original
loop consumed **every** uninitialized (`-`) entry from `git submodule status`,
not just the config-managed ones. In any repo with an unrelated uninitialized
submodule — a private Hugo theme, a vendored library, doc-example submodules —
the hook would try to clone it on **every** session start, which can block
startup or fail on credentials for a submodule `config` deliberately leaves
alone.

## Fix

Scope the bootstrap to **config-managed** submodules only — exactly the set
`./config/pull` floats:

- `config` itself (the bootstrap target; it carries no tracked `branch` in a
  consumer's `.gitmodules`, so it is added explicitly — mirroring `pull`'s
  step 1), plus
- every submodule that declares a tracked `branch` in `.gitmodules`
  (`.agents/shared`, and any shared submodule added later).

A new `config_managed_paths()` helper computes that set by parsing `.gitmodules`
with the same branch-tracking rule `pull` uses, so the two scripts can never
disagree about what is shared. The init loop now skips any `-` path that is not
in the set (noting the skip on stderr). Consumer-owned submodules are left
untouched.

`AGENTS.md` is updated to document the narrowed behavior.

## Verification

- `bash -n init-submodules` — clean.
- Computed managed set in this repo: `config`, `.agents/shared` — matches
  `pull`'s branch-tracked rule.
- Simulated `git submodule status` with mixed entries: `config` and
  `.agents/shared` → INIT; a Hugo theme and a vendored lib → SKIP; an
  already-initialized submodule → untouched.
- Ran `./init-submodules` live: clean no-op (the one submodule here is already
  initialized), exit 0.
- `review-docs` on the `AGENTS.md` change: APPROVE, no findings.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,9 +19,15 @@ Shared skills, scripts, and guidelines come from the `.agents/shared` submodule 
 `./config/pull` initializes and floats them automatically. But a fresh `git worktree`
 (and some shallow clones / cloud checkouts) start with NO submodules checked out, so
 those symlinks dangle and no skills are found. Bootstrap such a tree with
-**`./init-submodules`** — a root script that materializes the missing submodules
-(`config`, `.agents/shared`, …) at their pinned commits. It depends on no pre-existing
-`config` submodule, so it works before `./config/pull` (which lives inside the `config`
+**`./init-submodules`** — a root script that materializes the missing
+*config-managed* submodules at their pinned commits: `config` itself, plus every
+submodule that declares a tracked `branch` in `.gitmodules` (`.agents/shared`, and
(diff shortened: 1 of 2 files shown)


alexander-yevsyukov:
@codex review

armiol (approved):

PR #717, opened 2026-06-29
Bump Compiler and Logging

alexander-yevsyukov:


alexander-yevsyukov:
The dependency bumps will be addressed in another PR later.

PR #737, opened 2026-08-03
A `ModuleDependency` extension for excluding JetBrains Annotations

alexander-yevsyukov:
This PR adds the `ModuleDependency.excludeJetBrainsAnnotations()` extension which allows to exclude JetBrains annotation from the dependencies of a published artifact. This is needed to avoid the version conflict in the consumer projects.

### Other notable changes
 * Local dependencies were updated to the latest published versions.
 * The `ModelCompiler` dependency object was removed as no longer used.


armiol (approved):
@alexander-yevsyukov assuming the PR is "ready for review", LGTM.

